PR #17, opened 2021-03-12
Prepare release v0.4.0

owais:


src/version.ts
```diff
@@ -14,4 +14,4 @@
  * limitations under the License.
  */
 
-export const VERSION = '0.1.0';
+export const VERSION = '0.4.0';
```
jtmalinowski, line 17: Can you add an automated check in CI to prevent version mismatch?

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+## 0.4.0 (03-12-2021)
+
+### Changes
+
+- Changed environment variable prefix from `SPLK_` to `SPLUNK_`. All environment
```
jtmalinowski, line 9: This should trigger a major version bump right? It's only a minor bump because of leading 0 right?

src/version.ts
owais: I can create a ticket for it :)

jtmalinowski (approved):

CHANGELOG.md
owais: Correct. It should if we were 1.x but since we are still "unstable", we are allowed to break people's apps unfortunately for them :)

owais: I'd have added backward compatibility for this but I know only one customer is using this right now and I can communicate this change to them directly :)

PR #42, opened 2021-03-23
Update default propagators and allow overrides

owais:
Changes:

- Default propagator is now a composite propagator comprised of B3 and
tracecontext propagators.
- Options now accepts a `propagatorFactory` function that can be used to
provide a custom propagator.
- Replaced `provider.register()` which is not part of Otel API with
explicit calls to register provider and context manager.


src/options.ts
```diff
@@ -46,6 +51,7 @@ export interface Options {
   tracerConfig: NodeTracerConfig;
   spanExporterFactory: SpanExporterFactory;
   spanProcessorFactory: SpanProcessorFactory;
+  propagatorFactory: PropagatorFactory;
```
owais, line 54: I don't imagine propagators will need to be initialized lazily or will have any dependency on config or other components but for the sake of consistency I followed factory pattern.

package.json
```diff
@@ -79,6 +79,7 @@
     "@opentelemetry/propagator-b3": "^0.18.0",
     "@opentelemetry/resources": "^0.18.0",
     "@opentelemetry/tracing": "^0.18.0",
-    "jaeger-client": "^3.15.0"
+    "jaeger-client": "^3.15.0",
+    "semver": "^7.1.3"
```
owais, line 83: This is already an indirect dependency of @opentelemetry/node

src/options.ts
```diff
@@ -115,3 +124,13 @@ export function defaultSpanExporterFactory(options: Options): JaegerExporter {
 export function defaultSpanProcessorFactory(options: Options): SpanProcessor {
   return new BatchSpanProcessor(options.spanExporterFactory(options));
 }
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function defaultPropagatorFactory(options: Options): TextMapPropagator {
+  return new CompositePropagator({
+    propagators: [
+      new B3Propagator({ injectEncoding: B3InjectEncoding.MULTI_HEADER }),
```
seemk, line 132: Just to educate myself: is b3 multi the default propagator for Splunk distros? And why are both used?

src/tracing.ts
```diff
@@ -31,19 +35,25 @@ export function startTracing(opts: Partial<Options> = {}): void {
 
   _patchJaeger(options.maxAttrLength);
 
-  propagation.setGlobalPropagator(
-    new B3Propagator({ injectEncoding: B3InjectEncoding.MULTI_HEADER })
-  );
+  // propagator
```
seemk, line 38: nit: do we need these comments?

seemk (approved):

src/options.ts
iNikem: >is b3 multi the default propagator for Splunk distros

Yes

johnbley (approved):

src/options.ts
owais: > And why are both used?

In order to be compatible with both SFX libs and any other non-Splunk (or differently configured) Otel libs, and provide an easier migration path I'd say.

PR #46, opened 2021-03-24
More and better instrumentations

owais:
# Description

- Listed instrumentations as (optional) peer dependencies. This makes
require()'ing instrumentations safer despite @splunk/otel not listing
them as dependencies. Marking them optional ensures npm7 will not
automatically install these packages. Note that this will still result
in warnings for users on npm <7.
- Added suport for the following instrumentations out of the box:
  - @opentelemetry/instrumentation-express
  - @opentelemetry/instrumentation-ioredis
  - @opentelemetry/instrumentation-mongodb
  - @opentelemetry/instrumentation-mysql
  - @opentelemetry/instrumentation-net
  - @opentelemetry/instrumentation-pg
  - @opentelemetry/instrumentation-hapi
- Removed support for the following instrumentations:
  - @opentelemetry/hapi-instrumentation

## Type of change

# How Has This Been Tested?

- [x] Tested manually
- [x] Added automated tests

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,5 +81,61 @@
     "@opentelemetry/tracing": "^0.18.0",
     "jaeger-client": "^3.15.0",
     "semver": "^7.1.3"
+  },
+  "peerDependencies": {
+    "@opentelemetry/instrumentation-dns": "^0.14.0",
+    "@opentelemetry/instrumentation-express": "^0.14.0",
+    "@opentelemetry/instrumentation-graphql": "^0.14.0",
+    "@opentelemetry/instrumentation-grpc": "^0.18.0",
+    "@opentelemetry/instrumentation-hapi": "^0.14.0",
+    "@opentelemetry/instrumentation-http": "^0.18.0",
+    "@opentelemetry/instrumentation-ioredis": "^0.14.0",
+    "@opentelemetry/instrumentation-koa": "^0.14.0",
+    "@opentelemetry/instrumentation-mongodb": "^0.14.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -81,5 +81,61 @@
...
+    "@opentelemetry/instrumentation-express": "^0.14.0",
+    "@opentelemetry/instrumentation-graphql": "^0.14.0",
+    "@opentelemetry/instrumentation-grpc": "^0.18.0",
+    "@opentelemetry/instrumentation-hapi": "^0.14.0",
+    "@opentelemetry/instrumentation-http": "^0.18.0",
+    "@opentelemetry/instrumentation-ioredis": "^0.14.0",
+    "@opentelemetry/instrumentation-koa": "^0.14.0",
+    "@opentelemetry/instrumentation-mongodb": "^0.14.0",
+    "@opentelemetry/instrumentation-mysql": "^0.14.0",
+    "@opentelemetry/instrumentation-net": "^0.14.0",
```
seemk, line 96: Btw I think this is not published, or is it? At least I wasn't able to find it on npm

owais: It is but we were not including it before so it won't be on npm yet. The next release should have it.

owais: It is not. I think mysql is not either but including them is harmless as we don't try to auto install these. Things should work for users once these are published and users install them.

owais: Or do you prefer we remove these for now and issue a patch release later once they are published?

seemk (approved):

PR #93, opened 2021-05-04
Use gitlab for releases

owais:


flands (approved):
Remove RC when done please

PR #117, opened 2021-05-26
chore: pin deps to patch version, use @opentelemetry/api release candidate

seemk:
# Description

Use release candidate version of `@opentelemetry/api`.

All instrumentations need to use the same major version of OpenTelemetry API or context management will break (https://github.com/open-telemetry/opentelemetry-js-api/blob/4cda9bd10a8caff59f637e7d982387772c626866/src/internal/global-utils.ts#L71), causing instrumentations to use different API / context managers.

What's great is that some newer instrumentations (for example loggers) were published with a dependency on `1.0.0-rc`, while others remained on API `0.18.0`, causing context toe be lost between instrumentations.

This PR pins the versions to patch versions, so each of the instrumentation has a dependency on `1.0.0-rc` and nothing else, see these PRs for releases and versions:
https://github.com/open-telemetry/opentelemetry-js/commit/0c15999dccd377294a03cf7d36085216d6e8c5c1
https://github.com/open-telemetry/opentelemetry-js-contrib/commit/d9365c5d9aa543542a9b48cfe8bc59949101cb02
https://github.com/aspecto-io/opentelemetry-ext-js/commit/708790e6b00aeddc5f013f576a4a266936c379e9


owais:
So all instrumentations we are pinning now explicitly depend on API `1.0.0-rc.0` and everything should work with `1.0.0-rc.0`?

owais (approved):

seemk:
> So all instrumentations we are pinning now explicitly depend on API `1.0.0-rc.0` and everything should work with `1.0.0-rc.0`?

Yeah, the linked PRs (at least for OpenTelemetry) are the release PRs and the instrumentations there contain `^1.0.0-rc.0` for the API dependency, so I pinned the instrumentations to patch versions, since any newer releases, e.g. `0.20.0` for core or `0.17.0` for contrib might be breaking changes, changing the API dependency to `0.20.0`.

For example, a recent change to the core repo, switching from API `1.0.0-rc` to `0.20.0` for `http` instrumentation: https://github.com/open-telemetry/opentelemetry-js/commit/ac81e17c684c3eb9228a08669bb91fb28c406984#diff-7006c7070bcbe67e8dcc806e2baa00ede28dce729be2a23295d274a777f9aa6d

seemk:
Will open a new one with signed commits

PR #121, opened 2021-05-28
feat: add log correlation/injection

seemk:
# Description

Adds log injection support for `bunyan`, `pino` and `winston`. This option is disabled by default.

Not sure regarding the environment variable name, suggestions welcome.

Also fixed service name not being injected to provider resource.

Ticket ID: `APMI-1628`

## Type of change

# How Has This Been Tested?

- [x] Tested manually
- [x] Added automated tests

# Checklist:


package.json
```diff
@@ -97,6 +104,9 @@
     "@opentelemetry/instrumentation-net": "~0.16.0",
     "@opentelemetry/instrumentation-pg": "~0.16.0",
     "@opentelemetry/instrumentation-redis": "~0.16.0",
+    "@opentelemetry/instrumentation-bunyan": "~0.16.0",
+    "@opentelemetry/instrumentation-pino": "~0.16.0",
+    "@opentelemetry/instrumentation-winston": "~0.16.0",
```
owais, line 109: I think these should be added to `peerDependenciesMeta` section below as well.

src/instrumentations/index.ts
```diff
@@ -43,12 +43,24 @@ const supportedInstrumentations: [string, string][] = [
   ['opentelemetry-instrumentation-mongoose', 'MongooseInstrumentation'],
 ];
 
-export function getInstrumentations(): InstrumentationOption[] {
```
owais, line 46: Are these changes neccesary given `configureLogInjection` function also checks if log injection is enabled and skips configuration if not? We could remove these checks and always return all the instrumentations. If a user has installed one of these packages, it is likely they want log injection but even if they don't want it and it is installed, `configureLogInjection` will later just not configure these instrumentations.

src/instrumentations/logging.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  if (
+    typeof instrumentation['setConfig'] !== 'function' ||
+    typeof instrumentation['getConfig'] !== 'function'
+  ) {
+    return;
+  }
+
+  const logHook = (_span: Span, record: LogRecord) => {
+    record['service'] = options.serviceName;
```
owais, line 40: any reason to not have this in the instrumentation packages? Otel spec recommends correlating with resources https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/overview.md#log-correlation

seemk: AFAIK currently it's not possible to access the provider's resource information, or even reading span attributes; at least I haven't found any.

For reference:
https://open-telemetry.github.io/opentelemetry-js/interfaces/span.html
https://open-telemetry.github.io/opentelemetry-js/interfaces/tracerprovider.html

If that is the case I guess we need changes to the API to read this information, which might be useful for log correlation.

package.json
seemk: Thanks, fixed

src/instrumentations/index.ts
seemk: Good point, changed it to always load them.

And added a comment to the tests when checking for the amount of instrumentations loaded by default

Rauno56 (approved):

src/instrumentations/logging.ts
owais: Another option is to allow passing a resource directly to the instrumentation constructor. Users can do it manually and our distributions can do it automatically. Issue with this is that a single node project might want to use two different resources to represent sub-components which wouldn't work in this case.

seemk: Updated it to read from the SDK span resource attributes.

src/options.ts
```diff
@@ -87,7 +89,11 @@ export function _setDefaultOptions(options: Partial<Options> = {}): Options {
 
   const extraTracerConfig = options.tracerConfig || {};
   const tracerConfig = {
-    resource: new EnvResourceDetector().detect(),
+    resource: new EnvResourceDetector().detect().merge(
+      new Resource({
+        [ResourceAttributes.SERVICE_NAME]: options.serviceName,
```
owais, line 94: If service name is not provided via options but is provided with `OTEL_RESOURCE_ATTRIBUTES` or `OTEL_SERVICE_NAME` in future, will it get ovewritten by the value of `defaultServiceName` here?

seemk: The precedence is now `options.serviceName > env.SPLUNK_SERVICE_NAME > service.name from env.OTEL_RESOURCE_ATTRIBUTES > defaultServiceName`

owais (approved):

Issue #168, opened 2021-07-06
Link to migration guide is broken


jtmalinowski:
Link to "Migrate from the SignalFx Tracing Library for JS" [in the README](https://github.com/signalfx/splunk-otel-js/blame/main/README.md#L23) is broken.

jtmalinowski:
https://github.com/open-telemetry/opentelemetry.io/pull/626 is a blocker - we want the Open Telemetry registry to list all previously supported instrumentations.

jtmalinowski:
Notify https://github.com/signalfx/signalfx-nodejs-tracing/issues/78 when this issue is closed.

jtmalinowski:
https://github.com/open-telemetry/opentelemetry-js-api/issues/102 is a potential blocker.

PR #170, opened 2021-07-07
chore: deprecate SPLUNK_SERVICE_NAME

jtmalinowski:
Use OTel's `OTEL_SERVICE_NAME` instead of `SPLUNK_SERVICE_NAME`

# Description

Correcting this unnecessary divergence from the standard.

Also, needed for smoother migration process (#168)

## Type of change

# How Has This Been Tested?

- [x] Tested manually
- [x] Added automated tests

# Checklist:

### DIFF
--- a/test/tracing.test.ts
+++ b/test/tracing.test.ts
@@ -16,7 +16,6 @@
 
 import * as assert from 'assert';
 import * as sinon from 'sinon';
-import * as URL from 'url';
 import {
   BatchSpanProcessor,
   SimpleSpanProcessor,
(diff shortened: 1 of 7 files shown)


Rauno56 (approved):

test/tracing.test.ts
```diff
@@ -154,6 +157,8 @@ describe('tracing', () => {
       maxAttrLength,
       logInjectionEnabled
     );
+    stopTracing();
```
seemk, line 160: What was the reason `stopTracing` was added to tests? Here and at loginjection

seemk (approved):

test/tracing.test.ts
jtmalinowski: Well, so I made a mistake while coding, and I mistakenly thought that the tests were not passing because another test was overwriting the globals. However, I left `stopTracing` in place because deregistering is necessary for some tests in `signalfx/otel-js-web`, so I'm expecting that sooner or later it will be needed in this library too.

PR #172, opened 2021-07-09
fix: move packed artifact between stages

seemk:
`build` job did not publish anything, so produce the npm packed `.tgz` file from it to the release stage.

npm package was missing files due to `release` stage not having `lib` directory available after building


.gitlab-ci.yml
```diff
@@ -34,7 +34,7 @@ lint:
 build:
   artifacts:
     paths:
-      - dist/
+      - *.tgz
```
owais, line 37: Can we keep dist/ and move files to it in this job after running pack? It's be more consistent with other projects and I feel like building and moving packages to a dir should be in the same job.

.gitlab-ci.yml
```diff
@@ -50,9 +50,8 @@ release:
     - branches
   script:
     - mkdir dist
-    - npm pack
```
owais, line 53: Will artifacts from previous job (build) be automatically available in this job?

owais:
How about we just add the compile step to the release job just before running pack? The build job exists only to ensure that we can build the package after every commit/PR. We don't have to integrate it with release. 

I'm not sure if artifacts from build will be available automatically in release or if we need additional steps for that and this might need additional hit and trial. So I'm thinking to leave build as is and add the compile command to the release job.

owais:
But of course if you have already verified this or are very sure this is all we need, happy to merge this as well.

owais (approved):

PR #194, opened 2021-07-21
chore: migrate to GH Actions

jtmalinowski:
# Description

Replaces Circle CI with Github Actions.

Why? Circle CI is paid and only runs for users who are in the signalfx org. If we ever want to accept external contributions, we need to be able to run CI on them, and Circle is not suitable for that.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [x] Tested manually
- [x] Added automated tests

# Checklist:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+    branches:
+      - main
+permissions: read-all
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    permissions: read-all
+    strategy:
+      fail-fast: false
```
owais, line 14: Does this apply to each step for a nodejs version or the workflow as a whole? Makes sense to keep 12 running if 10 fails but do we really want to run compile if lint or test failed?

owais (approved):
LGTM. Suggest to allow to fail fast depending on what it refers to.

.github/workflows/ci.yml
Rauno56: > You can allow specific jobs in a job matrix to fail without failing the workflow run. For example, if you wanted to only allow an experimental job with node set to 15 to fail without failing the workflow run.
From: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

So it's about different node versions, not steps, if I understand correctly.

Rauno56 (approved):
If I'm right on Owais's comment then LGTM.

.github/workflows/ci.yml
jtmalinowski: Yeah it's about node versions, I agree in principle that a linting error could fail the whole matrix immediately and unit test in a specific Node.js versions shouldn't stop the matrix, but I don't think there's an easy way to do that.

PR #212, opened 2021-07-30
chore: add an example showcasing OT vs different OTel setups

Rauno56:
# Todo

- [x] Add README

# Description

This adds an example instrumenting a toy express app and sending traces to Splunk APM without a collector in the middle.

- APMI-1923

## Type of change

- [ ] This change requires a documentation update

# How Has This Been Tested?

- [x] Tested manually
- [ ] Added automated tests

# Checklist:

- [ ] Changelogs have been updated


Rauno56:
@theletterf, I just jotted down everything that the readme should, in my opinion, minimally include. I'm open to restructuring it or any other kind of feedback you might have.

theletterf (approved):
Let’s roll with it for now—we need a starting point.

examples/express/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "splunk-otel-example-express",
+  "private": true,
+  "version": "1.0.0",
+  "description": "Example of instrumenting a toy Express app in different ways",
+  "main": "server.js",
+  "scripts": {
+    "client": "env-cmd nodemon -r ./tracer.js client.js",
```
jtmalinowski, line 8: So this `-r` is smart, but let's highlight it in the README. There's just one `server.js` file and it's not immediately obvious how it's instrumented.

examples/express/tracer.jaeger.js
```diff
@@ -0,0 +1,31 @@
...
+const tracer = require('@splunk/otel').startTracing({
+	spanExporterFactory: (options) => {
+		const {
+			endpoint,
+			accessToken,
+		} = options;
+
+		return new JaegerExporter({
+	    endpoint: endpoint,
+		  username: 'auth',
```
jtmalinowski, line 27: is this because any username is required?

jtmalinowski:
@Rauno56 So I'm wondering if we shouldn't have separate directories side-by-side for OpenTelemetry vs. OpenTracing. As a newcomer, would I be able to quickly understand which line of code belongs to which system?

examples/express/tracer.jaeger.js
Rauno56: Not any. It literally has to be "auth".
https://github.com/signalfx/splunk-otel-js/blob/main/src/options.ts#L152-L153

Rauno56:
> @Rauno56 So I'm wondering if we shouldn't have separate directories side-by-side for OpenTelemetry vs. OpenTracing. As a newcomer, would I be able to quickly understand which line of code belongs to which system?

Not sure what would we gain by separating them if they are actually exactly the same - only thing changes is the file we require before running our app.
I think is beneficial to see that in fact, the two solutions are effectively enabled the same way to be convincing in encouraging people to migrate.
+ it would be a burden to keep them in sync.

examples/express/package.json
Rauno56: Rewrote the overview section.

examples/express/SETUP.md
```diff
@@ -0,0 +1,35 @@
+
+The ingest setup is for OpenTracing only and doesn't mention access tokens
+
```
jtmalinowski, line 3: Is there a reason for this many empty lines?

jtmal-signalfx (approved):
Yeah so let's merge it, and keep improving based on feedback we'll get.

PR #218, opened 2021-08-05
feat: replace Jaeger exporter with OTLP

Rauno56:
# Description

I'm making an assumption that Smart Agent does not support OTLP. Am I right on that?

- APMI-1932

## Type of change

- [ ] This change requires a documentation update

# How Has This Been Tested?

- [ ] Tested manually
- [x] Added automated tests

# Checklist:


Rauno56:
Converted it back to draft since Synthetic Run tests started failing. Probably has to do with updated deps and not the changed exporter.

MIGRATING.md
```diff
@@ -74,9 +74,8 @@ Rename environment variables:
 | ---------------------------------- | ------------------------------------ | ----- |
 | SIGNALFX_ACCESS_TOKEN              | SPLUNK_ACCESS_TOKEN                  |       |
 | SIGNALFX_SERVICE_NAME              | OTEL_SERVICE_NAME                    |       |
-| SIGNALFX_ENDPOINT_URL              | OTEL_EXPORTER_JAEGER_ENDPOINT        | if Jaeger is used |
-| SIGNALFX_ENDPOINT_URL              | n/a                                  | OTLP is not implemented yet |
-| SIGNALFX_RECORDED_VALUE_MAX_LENGTH | SPLUNK_MAX_ATTR_LENGTH               |       |
+| SIGNALFX_ENDPOINT_URL              | OTEL_EXPORTER_OTLP_ENDPOINT          |       |
```
owais, line 77: This is not a 1-1 mapping. We should add comment that this needs to be OTLP endpoint OR they need to use `OTEL_EXPORTER_JAEGER_ENDPOINT`

owais:
https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#opentelemetry-environment-variable 

We need to support `OTEL_TRACES_EXPORTER`. Check GDI spec for details ^

Rauno56:
> Converted it back to draft since Synthetic Run tests started failing. Probably has to do with updated deps and not the changed exporter.

It was a faulty test: https://github.com/signalfx/splunk-otel-js/pull/228

Rauno56:
> [https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#opentelemetry-environment-variable](https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md?rgh-link-date=2021-08-09T11%3A23%3A22Z#opentelemetry-environment-variable)
> 
> We need to support `OTEL_TRACES_EXPORTER`. Check GDI spec for details ^

Thanks for pointing that out! I see that something that needs to be added on top of the default, which this PR changes. Do you think we could add it in another PR?

MIGRATING.md
Rauno56: True, but aren't we asusming they're migrating to Splunk APM which, in a way, is an equivalent with OTLP-support?

Rauno56: I added https://github.com/signalfx/splunk-otel-js/pull/218/commits/7d3db551434cc4c063455d46494f1e01926b2700 for now, but rather remove it regarding my comment above.

owais:
> Thanks for pointing that out! I see that something that needs to be added on top of the default, which this PR changes. Do you think we could add it in another PR?

We could but it's better if such changes land together so we can review everything at once. May be land that feature first and then merge this PR? That might make it easier. 

Either way, totally up to you.

Rauno56:
> > Thanks for pointing that out! I see that something that needs to be added on top of the default, which this PR changes. Do you think we could add it in another PR?
> 
> We could but it's better if such changes land together so we can review everything at once. May be land that feature first and then merge this PR? That might make it easier.
> 
> Either way, totally up to you.

Since the feature of controlling the exporter via env var was not there before this PR, I'd merge this first and then make it easily configurable later. It makes more sense to me because the changes of this PR will need to be merged as is either way and the support for `OTEL_TRACES_EXPORTER` is additive to this - otherwise I'd need to possibly significantly change this PR later.

Please approve, if that sounds reasonable.

owais:
OK but you'll need to ensure no one else releases this before your next PR lands :)

src/tracing.ts
```diff
@@ -37,8 +36,6 @@ export function startTracing(opts: Partial<Options> = {}): void {
 
   const options = _setDefaultOptions(opts);
 
-  _patchJaeger(options.maxAttrLength);
```
owais, line 40: why are we removing this?

examples/express/.env.example
```diff
@@ -1,4 +1,4 @@
 SPLUNK_ACCESS_TOKEN='0123456789abcdefghijkl'
 OTEL_SERVICE_NAME='collectorless-express-example'
-OTEL_EXPORTER_JAEGER_ENDPOINT='https://ingest.us0.signalfx.com/v2/trace'
+OTEL_EXPORTER_OTLP_ENDPOINT='https://ingest.us0.signalfx.com/v2/trace/otlp'
```
owais, line 3: Do we use OTLP HTTP or gRPC exporter now and is it supported by SFX ingest?

owais (changes requested):
I took another look and it seems we are entirely dropping support for jaeger. That doesn't look right to me. I think this PR should just add support for OTLP exporter without removing any support for Jaeger.

examples/express/.env.example
Rauno56: We are using OTLP/HTTP which is supported by the endpoint listed :point_up:

Rauno56:
@owais, Jaeger support is now back behind `OTEL_TRACES_EXPORTER=jaeger-thrift-splunk` env var or 
`tracesExporter` config option.

examples/express/utils.js
```diff
@@ -20,6 +20,11 @@ const populateEnv = () => {
   if (!process.env.OTEL_EXPORTER_JAEGER_ENDPOINT) {
     process.env.OTEL_EXPORTER_JAEGER_ENDPOINT = process.env.SIGNALFX_ENDPOINT_URL;
   }
+  if (!process.env.OTEL_EXPORTER_OTLP_ENDPOINT) {
+    process.env.OTEL_EXPORTER_OTLP_ENDPOINT = (
+      process.env.SIGNALFX_ENDPOINT_URL || process.env.OTEL_EXPORTER_JAEGER_ENDPOINT
```
owais, line 25: I don't think we support SIGNALFX_ENDPOINT_URL in new Otel based libs anymore.

owais: I think we should remove these as the libs don't support these fallbacks and it might be confusing to keep them in examples.

src/options.ts
```diff
@@ -45,14 +45,21 @@ type SpanProcessorFactory = (
 
 type PropagatorFactory = (options: Options) => TextMapPropagator;
 
+export enum TracesExporter {
+  JAEGER = 'jaeger',
```
owais, line 49: should this be jaeger-thrift or jaeger_thrift?

src/options.ts
```diff
@@ -45,14 +45,21 @@ type SpanProcessorFactory = (
...
+  JAEGER_THRIFT_SPLUNK = 'jaeger-thrift-splunk',
+  OTLP = 'otlp',
```
owais, line 51: do we want this to be `otlp-grpc`?

src/options.ts
```diff
@@ -45,14 +45,21 @@ type SpanProcessorFactory = (
...
 export interface Options {
-  endpoint: string;
+  endpoint?: string;
   serviceName: string;
   accessToken: string;
   maxAttrLength: number;
   serverTimingEnabled: boolean;
   logInjectionEnabled: boolean;
   instrumentations: InstrumentationOption[];
+  tracesExporter: TracesExporter;
```
owais, line 62: we already have a spanExporterFactory option below, do we really need this?

src/options.ts
```diff
@@ -111,9 +113,29 @@ export function _setDefaultOptions(options: Partial<Options> = {}): Options {
...
   // factories
-  options.spanExporterFactory =
-    options.spanExporterFactory || defaultSpanExporterFactory;
+  if (options.spanExporterFactory === undefined) {
+    if (options.tracesExporter === TracesExporter.JAEGER) {
+      options.spanExporterFactory = jaegerSpanExporterFactory;
+      options.endpoint =
+        options.endpoint ||
+        otelEnv.OTEL_EXPORTER_JAEGER_ENDPOINT ||
+        'http://localhost:14268/v1/traces';
```
owais, line 128: options are passed as an argument to spanExporterFactory. Each factory should host code to parse/set defaults internally.

src/options.ts
```diff
@@ -111,9 +113,29 @@ export function _setDefaultOptions(options: Partial<Options> = {}): Options {
...
+  if (options.tracesExporter === undefined) {
+    options.tracesExporter = (process.env.OTEL_TRACES_EXPORTER ||
+      TracesExporter.OTLP) as TracesExporter;
+  }
+
   // factories
-  options.spanExporterFactory =
-    options.spanExporterFactory || defaultSpanExporterFactory;
+  if (options.spanExporterFactory === undefined) {
+    if (options.tracesExporter === TracesExporter.JAEGER) {
```
owais, line 123: I think this should be purely an env var, not an argument to the function as we already have spanExporterFactory argument.

src/options.ts
owais: To clarify, I mean this function should not have specific knowledge about each exporter. it should just pass on the options instance to the exporter factory. Then exporter factory should read it's relevant env vars, options.endpoint, etc.

src/options.ts
```diff
@@ -111,9 +113,29 @@ export function _setDefaultOptions(options: Partial<Options> = {}): Options {
...
+      options.spanExporterFactory = jaegerSpanExporterFactory;
+      options.endpoint =
+        options.endpoint ||
+        otelEnv.OTEL_EXPORTER_JAEGER_ENDPOINT ||
+        'http://localhost:14268/v1/traces';
+    } else if (options.tracesExporter === TracesExporter.JAEGER_THRIFT_SPLUNK) {
```
owais, line 129: I think jaeger and jaeger-thrift-splunk should have two different exporter factories each with their own default endpoints.

owais: Or may be a single factory that accepts the default endpoint as an argument.

src/options.ts
```diff
@@ -133,13 +155,23 @@ export function _setDefaultOptions(options: Partial<Options> = {}): Options {
...
     tracerConfig: tracerConfig,
+    tracesExporter: options.tracesExporter,
     spanExporterFactory: options.spanExporterFactory,
     spanProcessorFactory: options.spanProcessorFactory,
     propagatorFactory: options.propagatorFactory,
   };
 }
 
-export function defaultSpanExporterFactory(options: Options): JaegerExporter {
+export function defaultSpanExporterFactory(options: Options): SpanExporter {
```
owais, line 165: I think we should call this `otlpSpanExporterFactory` instead of default. The constructor decides what is default and what isn't. Factory itself has no business knowing if it is used as the default one.

owais (changes requested):
Overall direction looks good. Some superficial comments about public API, naming, etc.

examples/express/utils.js
Rauno56: This whole function is specifically for people migrating as the first preliminary step to be able to..
1. get something that looks like their old config up and running with OTEL;
2. see what needs to be converted and how.

src/options.ts
Rauno56: It's HTTP, ingress doesn't currently support grpc.
[The spec](https://github.com/signalfx/gdi-specification/blob/aea51ebe986de6980950ca3f7948d84420960d8b/specification/configuration.md) says that the default has to be `otlp`.

owais: Sorry I didn't mean gpprc specifically but just to add the format instead of just calling it otlp but if the spec recommends otlp then I suppose we use otlp as an alias for otlp-http.

examples/express/utils.js
owais: I don't think we ever had a goal to have config backward compatibility in Otel for old SFX libs. IMO this can lead to bugs as things will work with SIGNALFX config in this config but may not in another real project. IMO we should restrict the migration guide strictly to markdown documentation and not pollute otel examples with old sfx config. I'm sure this will lead to confusion. 

That said, I see you're just updating existing examples and didn't add new one so it's not a blocker for me for this PR. I still think we should remove references to SFX stuff later and keep pure otel examples here.

Rauno56: > I don't think we ever had a goal to have config backward compatibility in Otel for old SFX libs.

Not backwards compatibility, but helping through testing and conversion. I'd rather get some feedback from the first users, but totally open to removing this if it ever causes any confusion.

src/options.ts
Rauno56: Resolving this because it's the same argument as https://github.com/signalfx/splunk-otel-js/pull/218#discussion_r690266320

src/options.ts
Rauno56: Rest of `_setDefaultOptions` sets and parses defaults for other parts of the API and we already pass the factories the whole `options` object instead of letting it take some of it's configuration from `options` and rest from the env. I don't see how exporters are different.

src/options.ts
Rauno56: Since this is not in the spec, I'll make that more specific as you suggest using the key from OTel collector: `jaeger-thrift-http`

https://github.com/open-telemetry/opentelemetry-collector/blob/main/examples/k8s/otel-config.yaml

src/options.ts
Rauno56: I get where you are coming from, but I'd not add aliasing yet and leave only the minimal implementation per spec for now.

src/options.ts
Rauno56: We could live without it, but I think it's useful to have as much parity between different formats of configuration as possible. Sure, we cannot let people set functions through env vars which is why I'm open to removing this if you'd really prefer that and have it conceptually be a alternative to setting the function in "`config.js`" BUT through environment. WDYT?

src/options.ts
Rauno56: See https://github.com/signalfx/splunk-otel-js/pull/218#discussion_r693729544

src/options.ts
owais: I feel this is cluttering the constructor unnecessarily. Now we are providing two different ways of configuring an exporter with code. What if both are defined? Which one takes precedence? How do users decide which one to use when? If we accept tracesExporter as a well-known string, why not processor as well? What happens when we add metrics support? Do we add two options per component again?  This raises too many questions and doesn't solve enough problems (IMO) to be worth it. I think we should remove it and continue to use factory pattern.

src/options.ts
owais: What I meant was that the logic for configuring each exporter is leaking outside the factory to the constructor. I think constructor should do as little as possible aand only try to identify the components to be used. Then let the components configure themselves, including handling their own defaults, validation logic etc.

src/options.ts
Rauno56: Fair. Removed it. In doing so, introducing `otlp-http` alias became trivial, so I did that as well.

src/options.ts
Rauno56: See https://github.com/signalfx/splunk-otel-js/pull/218#discussion_r695641719

src/options.ts
Rauno56: Split the factories up.

src/options.ts
Rauno56: Changed that. Thanks.

src/options.ts
```diff
@@ -139,10 +134,57 @@ export function _setDefaultOptions(options: Partial<Options> = {}): Options {
...
+    SpanExporterMap[process.env.OTEL_TRACES_EXPORTER ?? 'default'];
+  assert.strictEqual(
+    typeof factory,
+    'function',
+    `Invalid value for OTEL_TRACES_EXPORTER env variable: ${util.inspect(
+      process.env.OTEL_TRACES_EXPORTER
+    )}. Pick one of ${util.inspect(Object.keys(SpanExporterMap), {
+      compact: true,
+    })} or leave undefined.`
+  );
```
owais, line 148: nit: IMO the following would be much simpler to read for most people including me but up to you if you want to change

```javascript

const factory = spanExporterMap[process.env.OTEL_TRACES_EXPRTER ?? 'default']
if  (factory == undefined) {
    throw new Error(...)
}
return factory
```

owais (commented):
Looks good. Left a couple of minor comments

src/options.ts
Rauno56: If you refer to the usage of `assert` then `if (...) { throw new Error(...); }` is exactly what that does - I do agree that it's somewhat less common outside of tests than if-error, but I do prefer that over if-errors if that's fine by you.
Or do you mean that the error message is long-winding?

src/options.ts
```diff
@@ -139,10 +134,57 @@ export function _setDefaultOptions(options: Partial<Options> = {}): Options {
...
+
+  if (options.accessToken) {
+    jaegerOptions.username = 'auth';
+    jaegerOptions.password = options.accessToken;
+  }
+
+  return new JaegerExporter(jaegerOptions);
+}
+
+export function splunkSpanExporterFactory(options: Options): SpanExporter {
```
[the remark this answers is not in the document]
Rauno56: I did not mind the duplication, but don't have a strong opinion on that. Changed it per your comment.

src/options.ts
owais: I don't mind the assert per se but I feel simple `if` condition to be much easier read and understand than a function that takes callbacks, type of object, a literal "function", etc. Anyone not an expert in JS/TS will have to spent a few minutes trying to understand it and probably read docs on assert function. On the other hand, if condition is universal and needs 5 second to understand even if you've never written JS before.

examples/express/utils.js
owais: I think we should have separate migration example and keep this one as pure Otel based but dont need it in this PR.

owais (approved):

owais:
@jtmalinowski please take a look as well. Thanks

src/options.ts
```diff
@@ -139,10 +134,40 @@ export function _setDefaultOptions(options: Partial<Options> = {}): Options {
...
+export function otlpSpanExporterFactory(options: Options): SpanExporter {
+  return new CollectorTraceExporter({
+    url: options.endpoint,
+    headers: {
+      'X-SF-TOKEN': options.accessToken,
+    },
+  });
+}
+
+export function genericJaegerSpanExporterFactory(
```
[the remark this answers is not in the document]
Rauno56: We do not. Removed the export. Thanks!

Rauno56:
@jtmalinowski gave his approval offline.

PR #219, opened 2021-08-05
chore: update basic example

Rauno56:
# Description

- Add README
- Update deps
- Add docs for running collectorless

# How Has This Been Tested?

- [x] Tested manually
- [ ] Added automated tests

# Checklist:

- [ ] Changelogs have been updated


jtmalinowski:
Looks good, 2 suggestions:
- add some basic e2e tests to make sure that examples actually work, add these tests to the CI pipeline too (I realise we'll do it anyway, because it's in the backlog, so up to you when)
- don't check package-lock in examples - this will better represent users' situation, where some unlocked external package might get updated and something breaks. We'll see it quickly once e2e tests are automated.

Neither of the above is a blocker.

Rauno56:
Thanks for the feedback! I'll remove the lockfile and create an issue for the e2e tests for now, not to get distracted from having this, the more extensive example and OTLP support on my table.

Rauno56:
> Neither of the above is a blocker.

I'll still need approval, @jtmalinowski. :pray:

jtmal-signalfx (approved):

PR #230, opened 2021-08-11
test: catch errors from propagation test

Rauno56:
# Description

One of the propagation tests would just time out if any of the assertions would fail.

## Type of change

# How Has This Been Tested?

- [ ] Tested manually
- [x] Added automated tests

# Checklist:

### DIFF
--- a/test/propagation.test.ts
+++ b/test/propagation.test.ts
@@ -24,18 +24,13 @@ import {
 } from '@opentelemetry/api';
 import { startTracing, stopTracing } from '../src/tracing';
 import { CompositePropagator, RandomIdGenerator } from '@opentelemetry/core';
-import {
-  InMemorySpanExporter,
-  ReadableSpan,
-  SpanProcessor,
-} from '@opentelemetry/tracing';
+import { InMemorySpanExporter, SpanProcessor } from '@opentelemetry/tracing';
 import { SYNTHETIC_RUN_ID_FIELD } from '../src/SplunkBatchSpanProcessor';
 import { defaultSpanProcessorFactory } from '../src/options';
 
 describe('propagation', () => {
-  it('must be set to b3', done => {
+  it('must be set to b3', () => {
     startTracing();
-
     assert(propagation.fields().includes('x-b3-traceid'));
     assert(propagation.fields().includes('x-b3-spanid'));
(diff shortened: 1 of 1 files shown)


test/propagation.test.ts
```diff
@@ -54,13 +49,11 @@ describe('propagation', () => {
       assert.strictEqual(carrier['x-b3-spanid'], spanId);
       assert.strictEqual(carrier['x-b3-sampled'], '1');
       assert.strictEqual(carrier['traceparent'], `00-${traceId}-${spanId}-01`);
-      done();
```
jtmalinowski, line 57: if for some reason this callback is not run then the asserts will never be run and we might get false positives

test/propagation.test.ts
```diff
@@ -102,15 +94,18 @@ describe('propagation', () => {
     const newContext = propagation.extract(context.active(), incomingCarrier);
     tracer.startSpan('request handler', {}, newContext).end();
 
-    spanProcessor.forceFlush().then(() => {
```
jtmalinowski, line 105: let's make the whole test function an async one and flatten this callback, like so:
```js
await spanProcessor.forceFlush();

assert(exporter.getFinishedSpans().length == 1);
// ...

stopTracing();
```

jtmalinowski (changes requested):
I added small suggestions to simplify things.

test/propagation.test.ts
Rauno56: By "callback" you mean `done`? If so the asserts in this test will be run regardless of us calling `done` or not. Do you have a code sample of a situation where the asserts would not be ran?

**Iff** the test function has a length of 1 (meaning we "ask" Mocha to make this an async test) `done` needs to be called though, if we don't add any arguments, Mocha assumes a synchronous test, which this is.

test/propagation.test.ts
Rauno56: I'll happily do that. Just didn't want to change more than I absolutely have to.

test/propagation.test.ts
jtmalinowski: By callback I mean the second argument to `context.with(spanContext, functionToRunInContext)`. I phrased it poorly.

jtmal-signalfx (approved):

PR #231, opened 2021-08-12
test: protect from environment leaking between tests

Rauno56:
# Description

Implemented an utility to clear relevant environment variables before running the tests.
Current ones are fortunately well-implemented and didn't have any errors regarding that, but
I stumbled upon fragile tests adding new in another PR.

## Type of change

# How Has This Been Tested?

- [ ] Tested manually
- [x] Added automated tests

# Checklist:

### DIFF
--- a/test/options.test.ts
+++ b/test/options.test.ts
@@ -37,8 +37,11 @@ import {
   Options,
   defaultPropagatorFactory,
 } from '../src/options';
+import * as utils from './utils';
 
 describe('options', () => {
+  beforeEach(utils.cleanEnvironment);
+
   it('verify default options', () => {
     // Mock the default `getInstrumentations` in case some instrumentations (e.g. http) are part of dev dependencies.
     const getInstrumentationsStub = sinon
--- a/test/resource.test.ts
+++ b/test/resource.test.ts
@@ -19,19 +19,12 @@ import * as sinon from 'sinon';
 
 import * as otel from '@opentelemetry/api';
 import { EnvResourceDetector } from '../src/resource';
+import * as utils from './utils';
 
 describe('resource detector', () => {
-  let envStub;
-
   beforeEach(() => {
+    utils.cleanEnvironment();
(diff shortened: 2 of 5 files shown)


test/options.test.ts
```diff
@@ -37,8 +37,11 @@ import {
   Options,
   defaultPropagatorFactory,
 } from '../src/options';
+import * as utils from './utils';
 
 describe('options', () => {
+  beforeEach(utils.cleanEnvironment);
```
jtmal-signalfx, line 43: let's clean after tests, not before, this suggests that we don't know where the "messy" environment settings come from

test/resource.test.ts
```diff
@@ -19,19 +19,11 @@ import * as sinon from 'sinon';
...
   beforeEach(() => {
-    process.env.OTEL_RESOURCE_ATTRIBUTES = '';
-  });
-
-  afterEach(() => {
-    if (envStub) {
-      envStub.restore();
-    }
-    delete process.env.OTEL_RESOURCE_ATTRIBUTES;
+    utils.cleanEnvironment();
```
jtmal-signalfx, line 26: again, let's clean after, this applies everywhere in the PR

jtmal-signalfx (approved):

test/options.test.ts
Rauno56: > suggests that we don't know where the "messy" environment settings come from
Well... we don't. And I feel bad for it honestly and would focus on making sure our tests run in a clean environment.

The environment is controlled by the runner. We don't assure the order of the tests executed so any of our tests might be the first, if your personal environment contains `OTEL_` prefixed stuff it already is affecting our tests - we should make sure we avoid falling into that.

Then... taking into account that we have to do before either way I don't see a reason to do it after as well.

jtmal-signalfx: Yeah if you think that's the right way to go then merge.

PR #232, opened 2021-08-16
feat: use w3c as the default propagation method

Rauno56:
# Description

Currently based on top of https://github.com/signalfx/splunk-otel-js/pull/218

- APMI-1933

## Type of change

# How Has This Been Tested?

- [ ] Tested manually
- [x] Added automated tests

# Checklist:


change/@splunk-otel-001c4603-3bb6-4d6f-8e53-fb10e4f5e929.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Make w3c(tracecontext, baggage) the default propagation mechanism",
+  "packageName": "@splunk/otel",
+  "email": "Rauno56@gmail.com",
```
seemk, line 5: What's this change directory for with the json files?

src/options.ts
```diff
@@ -58,6 +58,7 @@ export interface Options {
   spanExporterFactory: SpanExporterFactory;
   spanProcessorFactory: SpanProcessorFactory;
   propagatorFactory: PropagatorFactory;
+  propagators: string;
```
seemk, line 61: Should this perhaps be an array of `TextMapPropagator`? So when using the programmatic API (`startTracing({ ... })`) you'd be able to use a custom propagator.

change/@splunk-otel-001c4603-3bb6-4d6f-8e53-fb10e4f5e929.json
Rauno56: Was introduced in https://github.com/signalfx/splunk-otel-js/commit/527edb50b4a677832d868b5a4b54e8dce9f3099e. Check [CONTRIBUTING.md](https://github.com/signalfx/splunk-otel-js/blob/main/CONTRIBUTING.md#contributing-via-pull-requests).

src/options.ts
Rauno56: That's what `propagatorFactory` is for pretty much.
We actually went through a similar discussion with Owais on OTLP PR as well here: https://github.com/signalfx/splunk-otel-js/pull/218#discussion_r690267744
I'll give it another thought tomorrow and perhaps leave it as env-option only.

change/@splunk-otel-001c4603-3bb6-4d6f-8e53-fb10e4f5e929.json
seemk: Ah, thanks

src/options.ts
seemk: Ok I see, only thing that might be somewhat quirky then is that if programmatic API is used and a custom propagator factory is set which returns a FooPropagator and `propagators` is set as well (e.g. to `tracecontext`), which combination of final propagators is expected? I.e. we then have 2 ways to configure the same thing, but they conflict somewhat.

Anyway this seems to be turning into bikeshedding, guess we can live with it, purely a hypothetical case :smile:

seemk (approved):

src/options.ts
owais: Agree. I think we should stick with the factory only like we did in the other PR.

owais: > purely a hypothetical case 

I disagree Siim. These things confuse users and end up causing a lot of frustation and SWAT. No one wants more SWAT :)

seemk: Good point, perhaps leaving it env-only and the default factory just checks the env var?

Rauno56: Agree with both of you, **removed the option just like we did in OTLP PR**.
Sorry for that - I wrote this before I got any reviews on OTLP PR.

src/options.ts
```diff
@@ -226,3 +248,7 @@ function getEnvBoolean(key: string, defaultValue = true) {
 
   return true;
 }
+
+function deduplicate(arr: string[]) {
+  return [...new Set(arr)];
```
owais, line 253: nit: can this be used inline instead of wrapping in a single line, single use function?

owais (approved):

src/options.ts
Rauno56: It could be. The reason I extracted it is because I find this way to be more self-documenting than `[...new Set(propagatorsStr.split(',')))]` would.

Issue #266, opened 2021-09-03
Use the standard OTel Node SDK


Rauno56:
We are currently reimplementing some of the APIs that OTel components already provide and standardize which means the user has another one to learn and we're failing in complying with OTel standards.

We'd ideally provide the same API as NodeSDK OTel provides or just wrap the actual NodeSDK for the users' convenience, which would help with the documentation(familiar API), testing(those components are already tested in the wild) and development(coming from the OSS community).

PR #279, opened 2021-09-09
docs: update docs for the GA

theletterf:
- [x] Create a docs folder. Inside there should be:

  - [x] A page with all configuration settings relevant to our distro; separate them from the README if they're all there already (see https://github.com/signalfx/splunk-otel-java/blob/main/docs/advanced-config.md)
  - [ ] Are all settings configured via ENV vars? Or is there a runtime argument way, too, as with Java? Let's specify all ways of config, which is preferred, and why
  -  [ ] Instructions on how to correlate traces with logs, if that's possible (or a link to OTel docs on how to do that)
  -  [ ] A list of supported libraries and frameworks, or links to that info (see https://github.com/signalfx/splunk-otel-java/blob/main/docs/supported-libraries.md)
  -  [ ] Basic troubleshooting doc with things like enabling debug logging (see https://docs.splunk.com/Observability/gdi/get-data-in/application/java/troubleshooting/common-java-troubleshooting.html#nav-Troubleshooting)

Additionally:

- [x] VERSIONING.md file explaining how release happen (see https://github.com/signalfx/splunk-otel-java/blob/main/VERSIONING.md)
- [x] Clarify what features make our distro stand out compared to upstream?
- [ ] What are instrumentation plugins for? Can we present a use case?
- [ ] Remove `SPLUNK_MAX_ATTR_LENGTH`. Prefer the native OTel env vars instead which will be soon merged in upstream.


Rauno56:
> A page with all configuration settings relevant to our distro

All the current options are defined in README. See it in the code here: https://github.com/signalfx/splunk-otel-js/blob/main/src/options.ts#L49-L61 to double-check.

> Are all settings configured via ENV vars? Or is there a runtime argument way, too, as with Java? Let's specify all ways of config, which is preferred, and why

Most of the settings have both. The difference is only in options that do not make sense as an env var - when the thing you need to provide is a function or an object. `process.env.*` is the node way of accessing the environment <- a way to double-check that we've covered everything.

There's no preferred way I'd say. Env is easier to get off the ground with. The programmatic way gives you better control. I'd say it's always better to go with former **as long** as it's sufficient for your usecase.

> Instructions on how to correlate traces with logs, if that's possible (or a link to OTel docs on how to do that)

We auto-instrument popular logging libs, but you have to use one of those to get the correlation - which kinda makes sense, there's no reasonable way to know you're logging to stdout without those. The config options are documented, but no deeper context is given - we expect the user to know what "log injection" means and does. There are also [tests](https://github.com/signalfx/splunk-otel-js/blob/e3377427ec38b5253b4118dbf6ffb570605b2e2f/test/loginjection.test.ts) for those that might give you an idea how they work and what do expect.

> A list of supported libraries and frameworks

There's a list in the readme. Incomplete one though because we support everything that's officially supported by OTel and that list is constantly growing. See `opentelemetry-instrumentation-*` folders [here](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node) + [the OTel registry](https://opentelemetry.io/registry/?s=&component=instrumentation&language=js#).

> Basic troubleshooting doc with things like enabling debug logging

Enabling diagnostics logging is shown in [the express example](https://github.com/signalfx/splunk-otel-js/blob/main/examples/express/tracer.js).

> VERSIONING.md file explaining how release happen

We have https://github.com/signalfx/splunk-otel-js/blob/main/RELEASING.md.

> Clarify what features make our distro stand out compared to upstream

- We make it trivial to connect to Splunk's backends(wow, right?);
- Auto-load installed instrumentations;
- Bundle and configure best practices out of the box.
- Are ~80%* interoperable with standard OTel. See https://github.com/signalfx/splunk-otel-js/issues/266

\* I made this number up.

> What are instrumentation plugins for? Can we present a use case?

Just like for any other language distro they are an automatic way to generate library-specific spans. They give insight into the usage of the instrumented library with virtually 0 effort! See the examples.

theletterf:
@Rauno56 For GA, I think that the only docs that you need is adding a Troubleshooting section the README, which would benefit me greatly, too. Following what Python and Java did, I suggest that you also add a section on correlating traces and logs. For example, see https://github.com/signalfx/splunk-otel-java/blob/main/docs/correlating-traces-with-logs.md

I believe it is a requirement also to document all settings/config params: https://github.com/signalfx/gdi-specification/blob/main/specification/repository.md#data-collector. An example you'd follow is what the Java team did, with ENV vars and their runtime equivalents: https://github.com/signalfx/splunk-otel-java/blob/main/docs/advanced-config.md

Those pieces I can review and help improve, but the raw input to the repo must come from you. (I'd try and build the tables of settings based on what's in the README, but I might miss something. In any case, you should pull the settings from there and put them in a separate page.)

How does the above sound?

docs/advanced-config.md
```diff
@@ -0,0 +1,44 @@
...
+
+- Environment variables (also within `.env` files)
+
+   For example: `export OTEL_SERVICE_NAME='test-service'`
+
+- Arguments passed to the `startTracing()` function
+
+   For example: `startTracing({ serviceName: 'my-node-service', });`
+
+> Code arguments take precedence over the corresponding environment variables.
```
theletterf, line 13: @Rauno56 Is this right?

docs/advanced-config.md
```diff
@@ -0,0 +1,44 @@
...
+
+- `spanExporterFactory`: A function that accepts the options passed to the startTracing function. Returns a new instance of SpanExporter. When set, this function is used to create a new exporter and the returned exporter will be used in the pipeline.
+
+- `spanProcessorFactory`: Returns a SpanProcessor instance or an array of SpanProcessor instances. When set, this function is be used to create one or more span processors. The returned processors are added to the global tracer provider and configured to process all spans generated by any tracer provider by the global provider. The function creates a new span exporter and uses it with each processor it creates. It may call `options.spanExporterFactory(options)` to create a new exporter as configured by the user.
+
+- `propagatorFactory`: A function that returns a new instance of a TextMapPropagator. Defaults to a composite propagator comprised of W3C [Trace Context](https://www.w3.org/TR/trace-context/) and [Baggage](https://w3c.github.io/baggage/) propagators.
+
+- `instrumentations`: Can be used to enable additional instrumentation packages.
```
theletterf, line 44: @Rauno56 Why these aren't available as ENV vars? Could we make them available as ENV vars for GA or it doesn't make sense?

examples/express/README.md
```diff
@@ -81,9 +81,9 @@ There is an exmample [.env](./.env.collector) file included, which following com
 # make sure the collector is running
 docker run --name otel-collector -d -p 55681:55681 otel/opentelemetry-collector
 # run the example server
-npm run server:collecor
+npm run server:collector
 # in a separate terminal:
-npm run client:collecor
+npm run client:collector
```
theletterf, line 86: @Rauno56 Fixed a typo here.

README.md
```diff
@@ -15,10 +15,10 @@ The Splunk distribution of [OpenTelemetry JS](https://github.com/open-telemetry/
...
 This Splunk distribution comes with the following defaults:
 
-- [W3C context propagation](https://www.w3.org/TR/trace-context).
+- [W3C tracecontext and baggage propagation](https://www.w3.org/TR/trace-context).
 - [OTLP exporter](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-exporter-collector-proto)
-  configured to send spans to a locally running [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector)
-  (`http://localhost:55681/v1/traces`).
+  configured to send spans to a locally running [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector) over HTTP
+  (default endpoint: `http://localhost:55681/v1/traces`).
 - Unlimited default limits for [configuration options](#trace-configuration) to
```
theletterf, line 22: @Rauno56 Our JS distro sends data to the collector using OTLP over HTTP? If that's the case, the Access Token is mandatory and the default endpoint is always listening on :55681 instead of the usual gRCP port. Is that right? That's significantly different from others distros.

README.md
```diff
@@ -9,92 +9,70 @@
...
-3. Run node app with `-r @splunk/otel/instrument` CLI argument
+3. Run node app with the `-r @splunk/otel/instrument` CLI argument
 
 ```
 export OTEL_SERVICE_NAME=my-node-svc
 node -r @splunk/otel/instrument app.js
 ```
 
-That's it - the telemetry data is not sent to the locally running Opentelemetry Collector! You can also instrument your app with code as described [here](#instrument-with-code).
+That's it - the telemetry data is not sent to the locally running Opentelemetry Collector. You can also instrument your app with code as described [here](#instrument-with-code).
```
Rauno56, line 57:
```suggestion
That's it - the telemetry data is now sent to the locally running Opentelemetry Collector. You can also instrument your app with code as described [here](#instrument-with-code).
```

Rauno56:
```suggestion
That's it - the telemetry data is not sent to the locally running Opentelemetry Collector. You can also instrument your app programmatically as described [here](#instrument-with-code).
```
"with code" sounds childish to me. Can be easily convinced otherwise though.

docs/advanced-config.md
Rauno56: It really actually does not make sense to support ENV here.

1. The option expects constructors, not strings.
2. :point_up: could be changed to **also** support strings, but in that case, it would duplicate the list of installed instrumentations in `package.json`.

Rauno56 (approved):

PR #285, opened 2021-09-14
test: test examples

Rauno56:
# Description

This doesn't count as e2e tests yet, because it runs with a version of `@splunk/otel` that's already released and specified in the `package.json` of each of the examples.

**To have proper e2e tests, we'd like to use the current build in examples. Any thought about how it's best to be done?**

## Type of change

# How Has This Been Tested?

- [ ] Tested manually
- [x] Added automated tests

# Checklist:


.github/workflows/ci.yml
```diff
@@ -63,9 +59,26 @@ jobs:
...
+  integration:
+    runs-on: ubuntu-latest
+    permissions: read-all
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Get versions
+        run: |
+          docker-compose --version;
+          docker --version;
```
seemk, line 78: Could omit it?

test/examples/basic.docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+version: '3'
+services:
+  collector:
+    # hopefully 0.27.0 will include polling exporter
```
seemk, line 4: Current version is 0.35.0, is it available now?

test/examples/utils.js
```diff
@@ -0,0 +1,158 @@
...
+
+  return parent;
+};
+
+const waitSpans = (count, timeout = 8) => {
+  const collectorUrl = new URL(process.env.COLLECTOR_URL ?? 'http://localhost:8378');
+  collectorUrl.searchParams.set('count', count);
+  collectorUrl.searchParams.set('timeout', timeout);
+
+  return got(collectorUrl, {
```
seemk, line 92: It's possible to get spans from collector via a request? :exploding_head: Think I've overlooked that

seemk (commented):
To test the current version perhaps it'd be possible to package it as usual, then point package.json of examples to `"@splunk/otel": "file:..."` containing the built splunk-otel?

test/examples/express/snapshot.js
```diff
@@ -0,0 +1,455 @@
+// a console.log from a previous run
+module.exports = [
+  {
+    traceId: '2JXQNfGhIadZSxTgeG3joQ==',
```
jtmal-signalfx, line 4: This doesn't look right to me, the snapshot should only contain the data we assert and have control over. The easiest way to fix if we want to stick with this snapshot pattern would be to map OTel's Collector output to entries with just the attributes that are constant, and then `expect().toMatchSnapshot();`

.github/workflows/ci.yml
Rauno56: Could. I thought about that. Proved to be useful during my dev. Harms no one.

Do you think I should?

test/examples/basic.docker-compose.yml
Rauno56: I don't think so. The PR is still open: https://github.com/signalfx/splunk-otel-collector/pull/436

test/examples/utils.js
Rauno56: See the comment above about https://github.com/signalfx/splunk-otel-collector/pull/436. It's a custom exporter.

test/examples/express/snapshot.js
jtmal-signalfx: Second thing, overall I think that the trace is too long, if a test breaks there's quite a few spans to go through. Can we make it smaller and still test the behaviour we want to test?

test/examples/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:16-alpine
+
+USER node
+ENV TZ=UTC
+RUN mkdir /home/node/app
+WORKDIR /home/node/app
+
+COPY --chown=node:node ["package*.json", "*.lock", "./"]
+RUN npm install
```
jtmal-signalfx, line 9: `npm ci`?

test/examples/express/snapshot.js
Rauno56: That's exactly what I'm doing in `assertSpans` - ignoring the parts that are constant, asserting everything else.
I do not want to clean the snapshots up from the non-constant parts because

1. that's more code with next to no benefit.
2. that makes recreating those snapshots more complex and error-prone.
3. looses data(in the context of `traceId` it really doesn't matter, but in some cases it might).

test/examples/Dockerfile
Rauno56: No package-lock in the examples => no `npm ci`.

Rauno56:
@jtmal-signalfx, @seemk :pray: for your comments. I'm already experimenting with proper e2e tests in another branch - let me know if you have any other comments about testing examples specifically.

test/examples/Dockerfile
Rauno56: That lock file is for the test runner. I'm using the same Dockerfile for the examples as well as the test runner.

To use `npm ci` for the test runner(not a lot of benefits since there's like 1 dep) I should separate Dockerfiles and have a separate one for the test runner - not worth it imho.

test/examples/express/snapshot.js
jtmal-signalfx: I talked to Rauno over Slack, we're going forward with what we have right now.

jtmal-signalfx (approved):
I think there's still some issues with this, but it's a good start.

Rauno56:
Happy to iron them out in the future if you'd open the discussion for those! Thanks!

seemk (approved):

PR #304, opened 2021-09-23
test: introduce e2e testing for unpublished changes

Rauno56:
# Description

- APMI-1922

## Type of change

- [ ] Added automated tests


package.json
```diff
@@ -106,33 +106,33 @@
...
+    "@opentelemetry/instrumentation-pino": "~0.23.0",
+    "@opentelemetry/instrumentation-redis": "~0.23.0",
+    "@opentelemetry/instrumentation-winston": "~0.24.0",
+    "opentelemetry-instrumentation-amqplib": "~0.23.0",
+    "opentelemetry-instrumentation-aws-sdk": "~0.23.0",
+    "opentelemetry-instrumentation-elasticsearch": "~0.23.0",
+    "opentelemetry-instrumentation-kafkajs": "~0.23.0",
+    "opentelemetry-instrumentation-mongoose": "~0.23.0",
+    "opentelemetry-instrumentation-sequelize": "~0.23.0",
+    "opentelemetry-instrumentation-typeorm": "~0.23.0"
```
seemk, line 135: Why were these downgraded? :eyes:

Rauno56: Because when we install the local version we'd have conflicts in the examples. The peer dep versions are very restrictive right now.

The current solution is to update peer deps with the main package and used version in the examples.
An alternative solution would be to detect the local unpublished versions for peer deps and install those with the local compiled version of the distro package itself:

1. compile distro
2. read peer dep versions from package.json
3. per example do:
    1. see the new peer dep versions
    2. `npm install splunk-otel-js.tgz @opentelemetry/instrumentation-package@newversion`.

Rauno56: I'd be happy do implement the described solution in another PR.

seemk: But won't the downgraded peer deps cause issues for users? I.e. they'd do `npm install @opentelemetry/instrumentation-x`, get the 0.25.0 one and then get a bunch of warnings due to peer dep versions?
Could we perhaps start pinning them to something like `^0.23.0`?

Rauno56: Yes, maybe.
`^0.23.0` is the same as `~0.23.0` in 0.x, but I get what you mean - make it looser. I did think about it - but "how loose?" is the question.

From one side we'd practically want to have `* excluding anything that might break us` because, why not - instrumentations are a way to the fine control over the SDK.
On the other hand, we'd want to control the variability of the bundle in the wild to avoid solving issues with compatibility etc. In that extreme we would not care about people not being able to just npm install the latest, because we've got a list of instrumentation-version tuples we support and if they need others, they should upgrade to the later version of our package.

Thinking through all this during the development of this PR I concluded that it's outside of the scope for now. I'm turning back an unreleased change in this PR just to be able to go through that again explicitly with having the above in mind.

**My suggestion is still to leave it as is in this PR but, before the next release upgrade all instrumentation packages back to their latest versions, but hold them locked**. We can always loosen them once we get customer feedback, but it's more difficult to go the other way.

test/examples/basic/app.env
```diff
@@ -2,3 +2,4 @@ PORT=80
 OTEL_SERVICE_NAME='express-example'
 OTEL_RESOURCE_ATTRIBUTES='deployment.environment=dev'
 OTEL_LOG_LEVEL='DEBUG'
+OTEL_EXPORTER_OTLP_ENDPOINT=http://collector:55681/v1/traces
```
seemk, line 5: Is `55681` still the default for HTTP OTLP or is it `4318` now?

Rauno56: It's the default "at the time of writing this PR". I've finished another one that will take care of it https://github.com/signalfx/splunk-otel-js/pull/319

package.json
seemk: But wouldn't it best to have an up to date version that works with the latest OTel? Should get easier once 1.0 is reached, I'm just afraid we might make it too annoying for users to install anything. :shrug:

seemk: And the point about relaxing peer deps is fine, it's that we'd have an upper limit on the versions we are compatible with and in an ideal world we are always (as in same day) compatible with the latest OTel versions, unless they do a major breaking change.

seemk (approved):
I'm not very keen on the deps downgrade to get examples working, but I don't have a better solution for now, perhaps someone else would like a look as well

Rauno56:
> I'm not very keen on the deps downgrade to get examples working, but I don't have a better solution for now, perhaps someone else would like a look as well

As I mentioned - I definitely intend to turn those back before the next release.

Changing those now would break the e2e tests themselves because the examples would be installing incompatible versions of the peer deps. We cannot update the incompatible peer deps because `@splunk/otel` currently released locks those down. There's just no way I can solve it without making an exception for the current versions, which is a waste because, hopefully, the next version would have the peer deps more relaxed already.

.github/workflows/ci.yml
```diff
@@ -78,10 +78,26 @@ jobs:
...
+
+  e2e:
+    runs-on: ubuntu-latest
+    permissions: read-all
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Test basic example
+        working-directory: test/examples
+        run: docker-compose -f e2e.docker-compose.yml up --exit-code-from test
```
jtmal-signalfx, line 97: Can we run just the Collector in a separate container, and have test suite run from the base VM that GH Actions launch? For example see <https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.github/workflows/unit-test.yml>

test/examples/utils.js
```diff
@@ -85,13 +85,18 @@ const getParentSpan = (arr, span) => {
 };
 
 const waitSpans = (count, timeout = 11) => {
```
jtmal-signalfx, line 87: Can you convert this to something that uses a standardised test runner? I don't see anything in this file that a test runner couldn't handle. We're using `mocha` already, so it makes the most sense I think, although `jest` could handle matching of payloads.

jtmal-signalfx (changes requested):
Let's replace docker-compose with the behaviour built into GH Actions unless there's a good reason not to, it will simplify things. For example <https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.github/workflows/unit-test.yml#L16>.

test/examples/utils.js
Rauno56: I'd prefer to do that if we actually have more than one function to run and avoid premature optimization on that. I don't know how it would exactly be evolving for now and I don't see a point to take guesses. It's just a test setup afterall.

Rauno56:
> Let's replace docker-compose with the behaviour built into GH Actions unless there's a good reason not to, it will simplify things

I'm all for simplifying it! Could we get this merged for now and work on that once we don't have the race for the GA?

jtmal-signalfx:
I have removed my "request changes", if everyone else is happy with this PR, then go ahead and merge.

seemk:
If the docker-compose were to be moved to GH actions, wouldn't it make it impossible to run it locally?

Rauno56:
> If the docker-compose were to be moved to GH actions, wouldn't it make it impossible to run it locally?

If you mean "removed from GHA" then I do tend to agree.

One *could* totally make it work somehow.
The collector has to be spun up, the deps need to be installed for the tester as well as the example, the environments have to be cleaned for all those and we need to implement some of the functionality inside our tests(spinning up containers + the app before running the tests) - totally doable, but I don't think it would be less complex in the end.

PR #312, opened 2021-09-27
feat: remove injection options

Rauno56:
# Description

- POR-9619

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Added automated tests
- [x] Unit tests have been added/updated
- [ ] Documentation has been updated
- [x] Change file has been generated (`npm run change:new`)
- [ ] Delete this branch (after the PR is merged)

PR #317, opened 2021-09-28
feat: add the option to capture URL parameters

seemk:
# Description

Add the option to capture URL parameters from HTTP requests. Capturing can be done either by supplying a list of keys to capture (case-sensitive) or a custom function which gets invoked with the current span and query parameters.

Parameters are set as `http.request.param.${key}` span attribute.

## Type of change

# How Has This Been Tested?

- [ ] Tested manually
- [x] Added automated tests

# Checklist:

- [x] Delete this branch (after the PR is merged)

### DIFF
--- a/docs/advanced-config.md
+++ b/docs/advanced-config.md
@@ -44,4 +44,4 @@ The following config options can be set by passing them as arguments to `startTr
 
 - `instrumentations`: Can be used to enable additional instrumentation packages.
 
-- `captureRequestUriParams`: Either a list of keys (case-sensitive) of HTTP query parameters to capture or a function which gets invoked with the current span and query parameters to set a custom span attribute. When using the former, parameters are set as span attributes as `http.request.param.${key}`. Attribute keys are normalized at capture time, meaning `.` is replaced with `_` to avoid any attribute namespacing issues.
+- `captureRequestUriParams`: Either a list of keys (case-sensitive) of HTTP query parameters to capture or a function that gets invoked with the current span and query parameters to set a custom span attribute. When using the former, parameters are set as span attributes as `http.request.param.${key}`. Attribute keys are normalized at capture time, meaning `.` is replaced with `_` to avoid any attribute namespacing issues.
--- a/docs/advanced-config.md
+++ b/docs/advanced-config.md
@@ -44,4 +44,4 @@ The following config options can be set by passing them as arguments to `startTr
 
 - `instrumentations`: Can be used to enable additional instrumentation packages.
 
-- `captureRequestUriParams`: Either a list of keys (case-sensitive) of HTTP query parameters to capture or a function that gets invoked with the current span and query parameters to set a custom span attribute. When using the former, parameters are set as span attributes as `http.request.param.${key}`. Attribute keys are normalized at capture time, meaning `.` is replaced with `_` to avoid any attribute namespacing issues.
+- `captureHttpRequestUriParams`: Either a list of keys (case-sensitive) of HTTP query parameters to capture or a function that gets invoked with the current span and query parameters to set a custom span attribute. When using the former, parameters are set as span attributes as `http.request.param.${key}`. Attribute keys are normalized at capture time, meaning `.` is replaced with `_` to avoid any attribute namespacing issues.
--- a/docs/advanced-config.md
+++ b/docs/advanced-config.md
@@ -42,3 +42,5 @@ The following config options can be set by passing them as arguments to `startTr
 - `propagatorFactory`: A function that returns a new instance of a TextMapPropagator. Defaults to a composite propagator comprised of W3C [Trace Context](https://www.w3.org/TR/trace-context/) and [Baggage](https://w3c.github.io/baggage/) propagators.
 
 - `instrumentations`: Can be used to enable additional instrumentation packages.
+
+- `captureRequestUriParams`: Either a list of keys (case-sensitive) of HTTP query parameters to capture or a function which gets invoked with the current span and query parameters to set a custom span attribute. When using the former, parameters are set as span attributes as `http.request.param.${key}`. Attribute keys are normalized at capture time, meaning `.` is replaced with `_` to avoid any attribute namespacing issues.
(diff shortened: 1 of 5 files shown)


docs/advanced-config.md
```diff
@@ -42,4 +42,6 @@ The following config options can be set by passing them as arguments to `startTr
 
 - `propagatorFactory`: A function that returns a new instance of a TextMapPropagator. Defaults to a composite propagator comprised of W3C [Trace Context](https://www.w3.org/TR/trace-context/) and [Baggage](https://w3c.github.io/baggage/) propagators.
 
-- `instrumentations`: Can be used to enable additional instrumentation packages.
\ No newline at end of file
+- `instrumentations`: Can be used to enable additional instrumentation packages.
+
+- `captureRequestUriParams`: Either a list of keys (case-sensitive) of HTTP query parameters to capture or a function which gets invoked with the current span and query parameters to set a custom span attribute. When using the former, parameters are set as span attributes as `http.request.param.${key}`. Attribute keys are normalized at capture time, meaning `.` is replaced with `_` to avoid any attribute namespacing issues.
```
Rauno56, line 47:
```suggestion
- `captureRequestUriParams`: Either a list of keys (case-sensitive) of HTTP query parameters to capture or a function that gets invoked with the current span and query parameters to set a custom span attribute. When using the former, parameters are set as span attributes as `http.request.param.${key}`. Attribute keys are normalized at capture time, meaning `.` is replaced with `_` to avoid any attribute namespacing issues.

```

src/instrumentations/http.ts
```diff
@@ -14,13 +14,106 @@
...
+  return true;
+}
+
+function parseUrlParams(request: IncomingMessage) {
+  if (request.url === undefined) {
+    return {};
+  }
+
+  try {
+    // If Node <11 is supported, need to use the legacy API.
```
Rauno56, line 47:
```suggestion
    // As long as Node <11 is supported, need to use the legacy API.
```

src/instrumentations/http.ts
```diff
@@ -14,13 +14,106 @@
...
+  HttpRequestCustomAttributeFunction,
 } from '@opentelemetry/instrumentation-http';
-import { isSpanContextValid, TraceFlags } from '@opentelemetry/api';
+import { diag, isSpanContextValid, TraceFlags } from '@opentelemetry/api';
+import { Span } from '@opentelemetry/api';
+import * as Url from 'url';
+
+type IncomingHttpRequestHook = (span: Span, request: IncomingMessage) => void;
+
+function shouldAddRequestHook(options: Options): boolean {
```
Rauno56, line 30: I guess we're doing this to skip early? However, it decouples the actual hook creation from this check so we essentially need to do it again the hook creation function.

I'd suggest just use one place for the hook creation and return early there for optimization.
Also have a case which applies for invalid input.

src/instrumentations/http.ts
```diff
@@ -42,41 +135,55 @@ export function configureHttpInstrumentation(
...
+
+  if (shouldAddRequestHook(options)) {
+    const requestHook = createHttpRequestHook(options);
+    if (config.requestHook === undefined) {
+      config.requestHook = requestHook;
+    } else {
+      const original = config.requestHook;
+      config.requestHook = function (this: unknown, span, request) {
+        requestHook(span, request);
+        original.call(this, span, request);
```
Rauno56, line 184: Instead of always forcing the user to use ours, let's allow them to overwrite it if they need it instead.
See https://github.com/signalfx/splunk-otel-js/pull/312/files#diff-a41c4d0de1c425efb1ab1345aaadeaedc505a09298f326598d0124cce7296f89 for the idea.

Rauno56 (changes requested):

src/instrumentations/http.ts
seemk: Not sure about this, if we do this we leave out the possibility to add any customization points if we'd ever need access to the hooks.

src/instrumentations/http.ts
seemk: It's there just to avoid attaching a request hook in the first place. The other check is when attaching the internal request processor.

src/instrumentations/http.ts
Rauno56: If I understood you correctly we can still have customization points the same way, which makes me think I probably didn't.
Can you ellaborate?

seemk: For example if users have configured URL parameter capturing while also using a list of explicit instrumentations with their own hook:
```typescript
startTracing({
  captureRequestUriPrameters: ['sortBy'],
  instrumentations: [
    new HttpInstrumentation({
      requestHook: (span, req) => span.setAttribute('foo', process(req))
    })
  ]
})
```

The main issue is that `captureRequestUriPrameters` will be ignored unless they incorporate our `defaultRequestHook` into their hook. How should they know to do that?

Rauno56: I think we're clear now.

By documentation. If they're already fiddling with the advanced configuration, they must've read **some** docs. That's our opportunity to explain the tradeoffs.

As opposed to an advanced user not wanting the performance tradeoff or the additional attributes and not being able to turn it off at all. Since latter("this is not possible, sorry") is unrealistic to be a solution for forever, we eventually would need to introduce another env var `NO_REALLY_DO_NOT_ADD_REQ_HOOK=true`.

So we're talking about the tradeoffs between more (straightforward) control and being opinionated here. I'm leaning towards the former because that leaves us options to do what you suggest later, once we're convinced it's the most common usecase.

Rauno56: Sane defaults over magic! :tada:

seemk: But is it a sane default if you, as an user, have to know which hooks we are using internally? I think the real problem is that HTTP instrumentation allows to register only one hook not multiple.

Rauno56 (approved):
We had a discussion offline to clear everything out. Makes sense to me now! Thanks, @seemk!

seemk:
Renamed the option from `captureRequestUriParams` to `captureHttpRequestUriParams`

PR #318, opened 2021-09-28
test: fix import for HttpBaggagePropagator

seemk:
Fixes it for OTel JS 0.25.0, however in 0.26.0 this will be renamed to `W3CBaggagePropagator`.


Rauno56:
Why didn't the test fail?

Rauno56 (approved):

seemk:
> Why didn't the test fail?

To be honest I'm not sure, the propagatorFactory does not get invoked at all in the test

PR #319, opened 2021-09-28
feat: use OTLP gRPC exporter by default

Rauno56:
:exclamation: **Integration tests cannot work on this PR - those are using the latest published version of the package, which does not understand gRPC configuration which we are using for both e2e test variants.** :exclamation: 

# Description

Replacing OTLP/HTTP exporters with OTLP/gRPC(for the main default) and Jaeger(for collectorless) tandem.

- POR-9648
- POR-9649

## Type of change

- [x] Added automated tests


Rauno56:
@theletterf, could you please proofread the changed docs?

examples/express/README.md
```diff
@@ -9,20 +9,22 @@ In all the cases the application logic in  [server.js](./server.js) and [client.
 
 ## Setup
 
-Before starting install dependencies:
+Before starting install dependencies and run the collector for the examples which require it:
```
theletterf, line 12:
```suggestion
Before you start, install the dependencies and run the collector for the examples that require it:
```

examples/express/README.md
```diff
@@ -44,59 +46,61 @@ To start collecting and exporting tracing data from the app it requires
...
+3. SDK initialization before running the application.
 
-Once `.env` file is filled in with access token(see file [.env.example](./.env.example)) the app can be run by setting the environment variables and requiring SDK initialization just before the application code:
+All those steps are done for you in this example:
+
+1. Additional packages are already specified in the `package.json`.
+2. All the configuration is included in [an environment file `.env.collector`](./.env.collector).
+3. SDK intitialization in [tracer.js](./tracer.js).
+
+All that's left is to run it:
```
theletterf, line 57:
```suggestion
All the following steps are done for you in the example:

1. Additional packages are already specified in the `package.json` file.
2. All the configuration is included in [an environment file, `.env.collector`](./.env.collector).
3. SDK initialization happens in [tracer.js](./tracer.js).

All that's left is to run it:
```

examples/express/README.md
```diff
@@ -44,59 +46,61 @@ To start collecting and exporting tracing data from the app it requires
...
 
 ### Instrumented via OTel SDK, Jaeger Exporter
 
-There's an [tracer setup](./tracer.jaeger.js) to showcase replacing the default OTLP/HTTP exporter with one that exports in Jaeger format. The configuration is similar to the plain OTel SDK setup, but
-
-1. `@opentelemetry/exporter-jaeger` package has to be installed and the factory initiating the Exporter has to be passed to `startTracing`;
-2. The endpoint(`OTEL_EXPORTER_JAEGER_ENDPOINT`) for Splunk's APM has to be replaced by one that accepts Jaeger formatted tracing data: `https://ingest.<realm>.signalfx.com/v2/trace`.
+There's an [tracer setup](./tracer.jaeger.js) to showcase replacing the default OTLP/gRPC Exporter with one that exports in Jaeger format. The configuration is similar to the plain OTel SDK setup, but the Exporter(`OTEL_TRACES_EXPORTER`) has to be replaced by the default Jaeger Exporter which is bundled with the SDK: `jaeger-thrift-http`.
```
theletterf, line 70:
```suggestion
The [tracer setup file](./tracer.jaeger.js) showcases how to replace the default OTLP/gRPC Exporter with a Jaeger exporter. The configuration is similar to the default OTel SDK setup, but the exporter (`OTEL_TRACES_EXPORTER`) has to be replaced by the default Jaeger Exporter which is bundled with the SDK: `jaeger-thrift-http`.
```

examples/express/README.md
```diff
@@ -44,59 +46,61 @@ To start collecting and exporting tracing data from the app it requires
...
 npm run client:jaeger
 ```
 
-### Instrumented via OTel SDK, exporting to locally running collector
+### Instrumented via OTel SDK, exporting to Splunk APM using Jaeger Exporter
 
-Instead of sending telemetry data directly to Splunk APM API, one can also use OTel Collector to as a forwarder or an endpoint.
-There is an exmample [.env](./.env.collector) file included, which following commands use:
+Instead of using OTel Collector to forward the telemetry data to Splunk APM, one can also send it directly from the application.
+Once `.env` file is created(see file [.env.jaeger-splunk](./.env.jaeger-splunk) for an example) and the access token is substituted into the file. Run it using that:
```
theletterf, line 83:
```suggestion
Instead of using the OTel Collector to forward telemetry data to Splunk APM, you can also send it directly from the application. Once the `.env` file is created (see [.env.jaeger-splunk](./.env.jaeger-splunk) for an example) and the access token is replaced in the file, run the collector using the following commands:
```

examples/express/README.md
```diff
@@ -44,59 +46,61 @@ To start collecting and exporting tracing data from the app it requires
...
-# run the example server
-npm run server:collector
-# in a separate terminal:
-npm run client:collector
+npm run server
+# In a separate terminal:
+npm run client
 ```
 
+If unsure about the values to use for `.env` but familiar with configuring OpenTracing SDK, one can alternatively also use [.env.opentracing](./.env.opentracing) as the basis - the OTel configuration will be automatically derived from that in this example(see [utils.js](./utils.js) for the conversions).
```
theletterf, line 91:
```suggestion
If are not sure about the values to use for `.env`, but you're familiar with configuring the OpenTracing SDK, you can also use [.env.opentracing](./.env.opentracing) as the baseline configuration: The OTel configuration is automatically derived from the OpenTracing settings in the example (see [utils.js](./utils.js) for the conversions).
```

theletterf: @Rauno56 This is pretty cool. Could users use this method when migrating?

examples/express/README.md
```diff
@@ -44,59 +46,61 @@ To start collecting and exporting tracing data from the app it requires
...
 
 ```shell
 npm run server:opentracing
-# in a separate terminal:
+# In a separate terminal:
 npm run client:opentracing
 ```
 
-These commands require [SignalFX Tracing Library](https://github.com/signalfx/signalfx-nodejs-tracing)-compatible configuration in the `.env`(see the [example](./.env.opentracing.example)).
+These commands require [SignalFX Tracing Library](https://github.com/signalfx/signalfx-nodejs-tracing)-compatible configuration in the `.env`(see the [example](./.env.opentracing)).
```
theletterf, line 103:
```suggestion
These commands require a configuration in the `.env` file compatible with the [SignalFX Tracing Library](https://github.com/signalfx/signalfx-nodejs-tracing). See the [example](./.env.opentracing).
```

examples/express/README.md
Rauno56: Not... directly. The example also logs the configuration out so they will get an idea of the resulting configuration, but there's no "automatic" way implemented.

They could use the function from `utils.js` themselves though, but I think just reading the code will be informative enough for them to do the conversion once and for all.

Rauno56:
Nice! Thanks for all the copy suggestions, @theletterf!

package-lock.json
```diff
@@ -1,10534 +1,8 @@
 {
   "name": "@splunk/otel",
   "version": "0.12.0",
-  "lockfileVersion": 2,
```
seemk, line 4: Was it on purpose to downgrade the lockfile version?

package.json
```diff
@@ -90,10 +90,11 @@
     "winston": "3.3.3"
   },
   "dependencies": {
+    "@grpc/grpc-js": "^1.3.7",
```
seemk, line 93: Can we be sure future versions of `exporter-collector-grpc` will have a compatible version of `@grpc/grpc-js`? Is the dependency explicitly needed if the exporter depends on one?

Rauno56: The version of this should be updated with the exporter - this range is what was specified on the exporter.

The dep isn't explicitly needed, added it when lint pointed that out. Although, the only real reason to actually do that in our case is to avoid a situation where the exporter removes the dep which breaks us. Even if we do add the dep, it causes undefined behaviour because the exporter will not be compatible anymore.

Removing the dep and adding the lint exception makes more sense to me now - WDYT?

package-lock.json
Rauno56: It was not. Fixing.

package.json
Rauno56: See https://github.com/signalfx/splunk-otel-js/pull/319/commits/bfb7d93eb128628ed261e1b5f17a8bf768a09214.

theletterf (approved):

seemk (approved):

PR #375, opened 2021-10-28
feat: runtime metrics

seemk:
# Description

Add runtime metrics collection in a similar fashion as is done by [signalfx-nodejs-collect](https://github.com/signalfx/signalfx-nodejs-collect). However `signalfx-nodejs-collect` used 2 different native extensions to gather event loop and GC metrics. For event loop metric only the whole cycle duration (including the poll phase) was reported as a metric, which is quite useless. It should have measured event loop lag, which is the whole event loop cycle minus the time it took for IO, timer polling (i.e. no CPU used).  Besides this `signalfx-nodejs-collect` had a lot of dead code regarding events, which were never sent (e.g. on each GC cycle). So the decision was to drop `signalfx-nodejs-collect` for a custom native extension.

Currently the exporting step will still use the SignalFx client.

List of metrics exported (see [signalfx-nodejs-collect](https://github.com/signalfx/signalfx-nodejs-collect#metrics) for the previous list):

- `nodejs.memory.heap.total` (gauge, bytes) - Heap total via `process.memoryUsage().heapTotal`
- `nodejs.memory.heap.used` (gauge, bytes) - Heap used via `process.memoryUsage().heapUsed`
- `nodejs.memory.rss` (gauge, bytes) - Resident set size via `process.memoryUsage().rss`
- `nodejs.memory.gc.size` (cumulative_counter, bytes) - Memory collected by the garbage collector
- `nodejs.memory.gc.pause` (cumulative_counter, nanoseconds) - Time spent doing GC
- `nodejs.memory.gc.count` (cumulative_counter, count) - Number of times GC ran. (`signalfx-nodejs-collect`: nodejs.memory.gc.total)
- `nodejs.event_loop.lag.max` (gauge, nanoseconds) - Max event loop lag (`signalfx-nodejs-collect`: nodejs.event_loop.max)
- `nodejs.event_loop.lag.min` (gauge, nanoseconds) - Min event loop lag (`signalfx-nodejs-collect`: nodejs.event_loop.min)

Environment variables introduced:

- `SPLUNK_METRICS_ENABLED` (default: 'false') - Metrics are opt-in
- `SPLUNK_METRICS_ENDPOINT` (default: `http://localhost:9943`)
- `SPLUNK_METRICS_EXPORT_INTERVAL` (default: 5000)

Misc:

- Why `getSignalFxClient`? Custom metrics will still need to be supported. This PR allows to provide your own SignalFx client to `startMetrics`, but since we configure it with the same access token anyway, they can just use this helper method to access the currently used client.

## Type of change

# How Has This Been Tested?

- [x] Tested manually
- [x] Added automated tests

# Checklist:


docs/advanced-config.md
```diff
@@ -55,3 +55,22 @@ The following config options can be set by passing them as arguments to `startTr
...
 
 - `captureHttpRequestUriParams`: Either a list of keys (case-sensitive) of HTTP query parameters to capture or a function that gets invoked with the current span and query parameters to set a custom span attribute. When using the former, parameters are set as span attributes as `http.request.param.${key}`. Attribute keys are normalized at capture time, meaning `.` is replaced with `_` to avoid any attribute namespacing issues.
+
+### Metrics
+
+| Environment variable<br>``startMetrics()`` argument             | Default value           | Support | Notes
+| --------------------------------------------------------------- | ----------------------- | ------- | ---
+| `SPLUNK_METRICS_ENABLED`<br>`enabled`                           | `false`                 | Experimental | Enabled metrics export. See [metrics documentation](metrics.md) for more information.
+| `SPLUNK_METRICS_ENDPOINT`<br>`endpoint`                         | `http://localhost:9943` | Experimental | The SignalFx metrics endpoint to send to.
+| `SPLUNK_METRICS_EXPORT_INTERVAL`<br>`exportInterval`            | `5000`                  | Experimental | The interval, in milliseconds, of metrics collection and exporting.
```
theletterf, line 65: Perhaps an extra column for the `startMetrics` argument would make it easier to scan the table.

theletterf: s/SignalFx/Splunk

docs/advanced-config.md
```diff
@@ -55,3 +55,22 @@ The following config options can be set by passing them as arguments to `startTr
...
+
+#### Additional `startMetrics` config options
+
+- `signalfx`: A JS object with optional `client` and `dimensions` fields. If you have already setup a SignalFx client with custom configuration, it is possible use this for sending instead of creating, configuring a new one. `dimensions` object adds a pre-defined dimension for each datapoint. The format for `dimensions` is `{key: value, ...}`.
```
theletterf, line 69:
```suggestion
- `signalfx`: A JS object with optional `client` and `dimensions` fields. If you have already setup a Splunk client with custom configuration, you can use this for sending instead of creating, configuring a new one. `dimensions` object adds a pre-defined dimension for each datapoint. The format for `dimensions` is `{key: value, ...}`.
```

Can we rename the setting to "splunk"?

theletterf (changes requested):

owais:
How will this work with Otel? What happens when Otel introduces similar runtime metrics? Are we marking this as unstable for now so we have a path to migrate to Otel implementation later?

Or why aren't we contributing this to otel contrib?

seemk:
> How will this work with Otel? What happens when Otel introduces similar runtime metrics? Are we marking this as unstable for now so we have a path to migrate to Otel implementation later?
> 
> Or why aren't we contributing this to otel contrib?

Yes, it'll be unstable for now. Runtime metrics are disabled by default for that reason. When OTel semconvs and the metrics SDK is available, there will be a breaking change. And after semconvs are available we can start contributing the module upstream.

docs/advanced-config.md
seemk: Hmm, explicitly wanted to make it clear that we are dealing with [SignalFx Node.js client](https://github.com/signalfx/signalfx-nodejs). What do you think about removing the `signalfx` field completely? And replace it with 2 fields: `client` and `dimensions`?

owais:
OK. If it is experimental and users understand things can break then it's fine I think. That said, SignalFx never had a single library for metrics and traces. Customers always had to use two different libs. Why wouldn't we let customers use signalfx metrics lib till we have counterparts in otel ready?

docs/advanced-config.md
theletterf: The old library? Then it makes sense. Just wondering about backward compatibility if the switch to OTel metrics in the future (see Owais' comments).

theletterf:
I tend to agree with @owais here. What's the take of Ivo on this?

scripts/prebuild-current.js
```diff
@@ -0,0 +1,10 @@
+const prebuildify = require('prebuildify');
+
+prebuildify({
+  targets: [process.versions.node]
+}, err => {
+  if (err) {
+    console.error(err.message || err);
+    process.exit(1);
+  }
+});
```
Rauno56, line 10: Newline here please.

Rauno56 (approved):
I'd still rather see the native extension have a separate life-cycle in another package, but this works for now.

owais:
We should at least try to add missing conventions to Otel metrics semantic conventions. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md

ivomagi:
> I tend to agree with @owais here. What's the take of Ivo on this?

I think I can restate the goals

1. Satisfy the need for runtime metrics today, not when Otel metrics SDK is stable and semconvs available
2. Package the solution in a way where it is easy to adopt
3. Make sure the means to transition to Otel-world are taken into account and we know what is the upgrade path to Otel semconv and data model and SDK. There is an epic for next PI where this is taken apart in more details. 
4. Provide similar user experience to other instrumentation lis in this aspect (only Java so far)

Whether or not the macro requirements above can be fulfilled with one or two client-facing artifacts is engineering call. Maybe worth a call as well to go over the concerns.

src/metrics/index.ts
```diff
@@ -0,0 +1,308 @@
...
+        gcPauseMetric(registry, 'incremental_marking', info, timestamp),
+        gcPauseMetric(registry, 'process_weak_callbacks', info, timestamp),
+
+        gcCountMetric(registry, 'all', info, timestamp),
+        gcCountMetric(registry, 'scavenge', info, timestamp),
+        gcCountMetric(registry, 'mark_sweep_compact', info, timestamp),
+        gcCountMetric(registry, 'incremental_marking', info, timestamp),
+        gcCountMetric(registry, 'process_weak_callbacks', info, timestamp),
+      ]) {
+        cumulativeCounters.push(metric);
```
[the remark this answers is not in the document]
seemk: Thanks, didn't know about that!

test/examples/Dockerfile_test
```diff
@@ -0,0 +1,12 @@
...
+
+USER node
+ENV TZ=UTC
+RUN mkdir /home/node/app
+WORKDIR /home/node/app
+
+COPY --chown=node:node ["package*.json", "*.lock", "./"]
+RUN npm install
+
+COPY --chown=node:node ./ .
```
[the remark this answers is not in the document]
seemk: Added missing EOLs (and configured VS Code to emit them on saves)

docs/advanced-config.md
seemk: I added a link to the `SignalFx client` there and mentioned that metrics will change once OpenTelemetry becomes available.

docs/advanced-config.md
seemk: Added a link to SignalFx client. Followed the existing doc style and `startTracing` did not have the extra column either, should I split both of them to separate column?

theletterf (approved):

PR #385, opened 2021-11-11
CI: add prebuilds

seemk:
- Upgraded the supported version to Node <18
- Prebuilds: done on a combination of OS(Linux, macOS, Windows) x Node(8.5, 9.0, 10.0, 11.0, 12.0, 13.0, 14.0, 15.0, 16.0, 17.0)
- The NPM package tgz is created in github, stored as an artifact
- GitLab fetches the artifact and publishes the package


scripts/prepare-release-artifact.js
```diff
@@ -0,0 +1,123 @@
...
+
+  if (run === undefined) {
+    throw new Error(`Workflow not found for commit ${commitSha}`);
+  }
+
+  return run;
+}
+
+async function waitForWorkflowRun(context) {
+  const waitUntil = process.hrtime.bigint() + BigInt(WORKFLOW_TIMEOUT_MS) * 1_000_000n;
```
Rauno56, line 34: Don't think we need such precision for that. `Date.now()` would be plenty accurate.

scripts/prepare-release-artifact.js
```diff
@@ -0,0 +1,123 @@
+const fs = require('fs');
+const { execSync } = require('child_process');
+const path = require('path');
+const { Octokit } = require('octokit');
+
+const { version } = require('../package.json');
+
+const WORKFLOW_TIMEOUT_MS = 15 * 60 * 1000;
+
+async function sleep(ms) {
```
Rauno56, line 10: This doesn't need to be an async function just return a promise, don't await it.

Rauno56 (approved):

scripts/prepare-release-artifact.js
seemk: fixed

PR #401, opened 2022-01-06
[DRAFT] Profiling

seemk:
Doing CI runs


src/profiling/profiling_contextmanager.ts
```diff
@@ -0,0 +1,65 @@
...
+      recorder = {
+        enterContext: () => {},
+        exitContext: () => {},
+      };
+    }
+
+    this._recorder = recorder;
+
+    this._enterContextOriginal = this['_enterContext'];
+    this['_enterContext'] = this._enterContextOverride;
```
Rauno56, line 43: Why not override the method and call `super.enterContext()` in there instead?

src/profiling/types.ts
```diff
@@ -0,0 +1,58 @@
...
+  /** Formatted stacktrace. */
+  stacktrace: string;
+  traceId: string;
+  spanId: string;
+}
+
+export interface ProfilingData {
+  /** Timestamp when profiling was started (nanoseconds since Unix epoch). */
+  startTimeNanos: string;
+  stacktraces: ProfilingStackTrace[];
```
Rauno56, line 35: Casing difference between `stacktraces` and `ProfilingStackTrace` either consider `stackTraces` or `ProfilingStacktraces` or just call them "stacks" like ECMAscript does.

Rauno56 (commented):
:+1: Awesome work. The cpp parts are mostly magic for me.
 
There was a discussion between us about moving compiled resources into a separate package after the implementation of the metrics.
IIRC you were looking into introducing another repo?
Now there's even more of that and I feel even stronger separating those concerns.

src/instrument.ts
```diff
@@ -16,6 +16,11 @@
 
 import { startTracing } from './tracing';
 import { startMetrics } from './metrics';
+import { startProfiling } from './profiling';
+import { getEnvBoolean } from './options';
 
+startProfiling({
+  enabled: getEnvBoolean('SPLUNK_PROFILER_ENABLED', false),
```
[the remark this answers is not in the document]
seemk: The main issue is that both metrics and profiling need to be switched on via an environment variable and are exported as functions.

When calling `startProfiling` programmatically (i.e. advanced config), setting the `enabled` option would make things ugly.

```javascript
startProfiling({ serviceName: 'foo', enabled: true }); // ??? I'm already calling start, why do I need to set it enabled
startProfiling({ serviceName: 'foo' }); // That's the way it should work, no extra params necessary.
```
The second option is the one I opted for, so it would work when starting instrumentation via CLI options as well.

I agree the enabled flag should be removed and `instrument.ts` just checks the env var without passing the flag to `startProfiling`, hoisted the flag out in https://github.com/signalfx/splunk-otel-js/pull/401/commits/f03ca4bd7b086096e86e33309c8884ff3e7bd545

src/profiling/profiling_contextmanager.ts
seemk: `_enterContext` is private. The other option would be to copy the context manager code here, not sure which is the best option for us

src/profiling/types.ts
seemk: Ah thanks, I initially typed everything as `stackTrace`, but then changed to `stacktrace` and missed the type casings.

seemk:
> +1 Awesome work. The cpp parts are mostly magic for me.
> 
> There was a discussion between us about moving compiled resources into a separate package after the implementation of the metrics. IIRC you were looking into introducing another repo? Now there's even more of that and I feel even stronger separating those concerns.

My current thinking is to keep them in the same codebase (feel free to sway me in either direction).

Points for the native extension being in a separate package:
* CI is a bit faster
* Codebase is slightly smaller to navigate in

Points against the native extension being in the same package:
* More maintenance overhead (maintaining 4 different CIs instead of 2 due to the repo mirroring we have :weary:)
* Slower feedback loop when developing (build the native package, then build this package and then do the test runs)

At the moment I haven't put it into a separate repo just because it's less work to do so.

package.json
```diff
@@ -44,9 +44,10 @@
   "files": [
     "binding.gyp",
     "prebuilds",
+    "protos",
     "index.js",
-    "src/native_ext/*.h",
-    "src/native_ext/*.cpp",
```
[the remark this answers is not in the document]
seemk: They are, but the native sources need to be included in the package, because if some combination of the prebuilt libraries doesn't exist there will be an attempt to build from source.

docs/profiling.md
```diff
@@ -0,0 +1,24 @@
...
+export SPLUNK_PROFILER_ENABLED=true
+node -r @splunk/otel/instrument app.js
+```
+
+Or when using [advanced configuration](advanced-config.md):
+
+```javascript
+const { startProfiling } = require('@splunk/otel');
+
+startProfiling({
```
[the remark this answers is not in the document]
seemk: For now yes, will be fixed once we get to beta

theletterf (approved):

Rauno56 (approved):

PR #412, opened 2022-02-07
doc: update instrumentations list

seemk:
Add the missing instrumentations that are loaded automatically in https://github.com/signalfx/splunk-otel-js/blob/main/src/instrumentations/index.ts

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ By default the following instrumentations will automatically be enabled if insta
 any of these instrumentations, you'll need to install them with npm and then run your app with `-r @splunk/otel/instrument` flag as described above.
 
 ```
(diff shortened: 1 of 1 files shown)


theletterf (approved):

Rauno56:
Could we just refer them to officially supported instrumentations instead of regularly updating this doc, @seemk ?

seemk:
> Could we just refer them to officially supported instrumentations instead of regularly updating this doc, @seemk ?

Think it needs to be in sync with the autoloader?

seemk:
Anyway we can discuss it later, merging it for now since we just stumbled upon the incomplete list

PR #420, opened 2022-02-21
feat: add env var to incorporate redis command's arguments

seemk:
New boolean env var `SPLUNK_REDIS_INCLUDE_COMMAND_ARGS`.

If set to `true`, we'll use a custom `dbStatementSerializer` for `redis` instrumentation to include the command arguments. The default logic only included the command name.

`db.statement` attribute will become:
when the flag is disabled (by default): `GET foo` -> `GET`
when enabled: `GET foo` -> `GET foo`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,7 +101,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu-latest', 'windows-latest', 'macos-latest']
+        os: ['ubuntu-latest', 'windows-2019', 'macos-latest']
         nodejs: ['10', '12', '14', '16', '17']
     steps:
       - name: Checkout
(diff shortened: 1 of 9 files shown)


seemk:
> 1. Not sure if features like these need to absolutely be included in our distro.
> 2. Additionally, the serialization might be more naive than expected:
> 
> ```
> assert.strictEqual(
>   `${[{ a: 123, }, null, undefined, () => { return 1; }].join(' ')}`,
>   '[object Object]   () => { return 1; }'
> );
> ```
> 
> 3. A test would be nice.

1. What's the alternative?
2. In these cases it's either an argument error or the redis client will just convert the args to strings in the same fashion. E.g. `() => {}` will actually appear in redis in the same form. I've only added buffer sanitization here to avoid arbitrary binary data in attribute value strings.

mhennoch (approved):

seemk:
Doing a new PR, CI is borked

PR #445, opened 2022-04-21
chore: remove jaeger-client

seemk:
Not directly used in our code. OTel Jaeger exporter still brings in the package however.


jtmal-signalfx (approved):

jtmal-signalfx (changes requested):
Let's pause for a second, and double check if we need any docs changes

jtmal-signalfx (approved):

PR #447, opened 2022-04-21
refactor: update env detector from upstream

Rauno56:
# Description

No significant updates. Updated the env detector from upstream to make the next PR(adding process detector) smaller.

## Type of change

# How Has This Been Tested?

- [x] ~~Added~~Updated automated tests

# Checklist:

### DIFF
--- a/src/resource.ts
+++ b/src/resource.ts
@@ -79,7 +79,7 @@ export class EnvDetector {
         const parsedAttributes = this._parseResourceAttributes(rawAttributes);
         Object.assign(attributes, parsedAttributes);
       } catch (e) {
-        diag.debug(`EnvDetector failed: ${e.message}`);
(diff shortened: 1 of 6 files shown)


src/resource.ts
```diff
@@ -118,7 +162,17 @@ export class EnvResourceDetector {
...
+   *
+   * @param str The String to be validated.
+   * @returns Whether the String is valid and not empty.
+   */
   private _isValidAndNotEmpty(str: string): boolean {
     return str.length > 0 && this._isValid(str);
   }
 }
+
+export const envDetector = new EnvDetector();
```
seemk, line 177: Is this used anywhere?

Rauno56: Nope, not yet. Straight from upstream.
Makes sense on the other hand because it has no internal state.

Rauno56: I'll merge for now, we can continue the discussion.

PR #448, opened 2022-04-21
feat: add ProcessDetector

Rauno56:
# Description

Adding a sync verison of [ProcessDetector](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-resources/src/detectors/ProcessDetector.ts) and an utility function that detects all listed(Env and Process for now) resources and merges them together. Swapping everything to use the "combined detector" for now with no configurability. An example of process attributes:

```
  +       'process.executable.name': '/[ ... ]/.nvm/versions/node/v16.13.2/bin/node',
  +       'process.pid': 432605,
  +       'process.runtime.description': 'Node.js',
  +       'process.runtime.name': 'nodejs',
  +       'process.runtime.version': '16.13.2',
```

This means that we now also have runtime version among the telemetry data which we didn't use to by default.

## Type of change

# How Has This Been Tested?

- [x] Added automated tests

# Checklist:


src/detectors/EnvDetector.ts
```diff
@@ -0,0 +1,182 @@
...
+   *
+   * @param str The String to be validated.
+   * @returns Whether the String is valid and not empty.
+   */
+  private _isValidAndNotEmpty(str: string): boolean {
+    return str.length > 0 && this._isValid(str);
+  }
+}
+
+export const envDetector = new EnvDetector();
```
seemk, line 181: If the detectors have no state, couldn't we just make them free functions?

seemk (approved):

src/detectors/EnvDetector.ts
Rauno56: Why would we implement detectors differently than OTel has it?

PR #460, opened 2022-05-05
chore: allow gitlab prereleases

seemk:
Tags with name matching `prerelease-v[0-9]+\.[0-9]+\.[0-9]+(?:-.+)?$`, e.g. `prerelease-v0.21.0` can be used to publish to npm. GitHub releases are skipped to avoid polluting releases list with experimental releases.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,25 +4,27 @@ default:
 stages:
   - release
 
+.prepare_dist: &prepare_dist
+  - npm install
+  - npm run prepare-release-artifact
+  - shasum -a 256 dist/* > dist/checksums.txt
+
+.release_npm: &release_npm
+  - echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/.npmrc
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -25,3 +23,20 @@ release:
...
     - npm publish ./dist/splunk-otel-${CI_COMMIT_REF_NAME:1}.tgz
     - rm -f ~/.npmrc
+
+prerelease:
+  stage: release
+  artifacts:
+    paths:
+      - dist/
+  rules:
+    - if: $CI_COMMIT_BRANCH =~ /^prerelease-.+/ && $CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+\.[0-9]+-.+/
```
Rauno56, line 33: This means only tags on a specific branch will have this flow ran?

seemk: Changed the yaml, the trigger is only based on tag names now. So it will be possible to create feature branches and publish experimental versions from there

PR #461, opened 2022-05-06
chore: remove branch check from prepare-release-artifact

seemk:
`CI_COMMIT_BRANCH` is not available on tag pushes anyway and the option would always default to `main`. This is needed for the prerelease tag pushing, so feature branches can be used.


Rauno56 (approved):
Just to double-check - we still check a certain pattern of the tag, right?

seemk:
> Just to double-check - we still check a certain pattern of the tag, right?

The pattern is checked in GitLab + the release script only works on protected tags. I added prerelease tag pattern to protected tags.

PR #464, opened 2022-05-11
feat: add instrumentations

Rauno56:
# Description

There are new instrumentations in OTel now. Add those to be autoloaded.

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation change or requires a documentation update

# How Has This Been Tested?

- [ ] Tested manually
- [ ] Added automated tests

# Checklist:

- [ ] Unit tests have been added/updated
- [ ] Documentation has been updated
- [ ] Change file has been generated (`npm run change:new`)
- [ ] Delete this branch (after the PR is merged)

PR #465, opened 2022-05-11
test: do not leak the sinon stub to other tests on failure

Rauno56:
# Description

if `loader silently fails when package is not installed` test fails, it brings down others and creates confusing output. Make sure the stub is removed even if the test fails.

## Type of change

# How Has This Been Tested?

- [x] Added automated tests

# Checklist:


Rauno56:
This caused the confusion in https://github.com/signalfx/splunk-otel-js/pull/464

seemk (approved):

PR #471, opened 2022-05-12
feat: make loading instrumentations npm-link-safe

Rauno56:
# Description

When `@splunk/otel` is npm linked, instrumentation loader cannot find the instrumentation packages installed in the main module and causes issues hard to debug.

This PR makes it so the loader first tries to load the instrumentations form the perspective of the main module(the expected behavior) and then falls back to the naive require which happens to behave the same way when `@splunk/otel` is not linked.

Another scenario that it fixes is when your dep tree looks something like this:

```
your/app
├─┬ metapackage
│ ├── instrumentation1
│ └── instrumentation2
└── @splunk/otel
```

Since `require` is called from the context of `@splunk/otel` none of the instrumentations are present from the context of our distro. Only if npm flattens the tree, they are.

## Type of change

# How Has This Been Tested?

- [x] Tested manually
- [x] Added automated tests

# Checklist:

### DIFF
--- a/change/@splunk-otel-7e8fbff4-a7ff-4cb3-97dc-70625c2a91e5.json
+++ b/change/@splunk-otel-7e8fbff4-a7ff-4cb3-97dc-70625c2a91e5.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "feat: make loading instrumentations npm-link-safe",
+  "packageName": "@splunk/otel",
+  "email": "rauno56@gmail.com",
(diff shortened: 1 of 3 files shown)


Rauno56:
As much as I'd like to just fix and forget it, it's probably not worth it for the rare use-cases it solves and can be bypassed via `--preserve-symlinks` + testing this I realized there will most likely be other problems on our way to make it work seamlessly anyway.
Thanks, @jtmal-signalfx, for thinking it through with me.

PR #473, opened 2022-05-17
doc: update profiling API usage

seemk:


Rauno56 (approved):
This will not be "fixed" really, will it? It will always be the case, we'll just make it more convenient probably in the future. I don't mind the wording for now though.

PR #474, opened 2022-05-19
feat: add a generic `start` API

Rauno56:
# Description

Could be merged roughly as is, but resource detection should be shared between the signals.

I currently documented `start` to enable all stable signals. That means stabilizing signals is a breaking change. **Are we OK with that?**. Alternative would be to:

1. Enable nothing by default or
2. Only ever enable tracing by default - similar to right now, but more explicit.

I would not want to add another option to toggle that behavior.

## Type of change

# How Has This Been Tested?

- [ ] Tested manually
- [x] Added automated tests

# Checklist:


src/start.ts
```diff
@@ -0,0 +1,63 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { getEnvBoolean } from './utils';
+import { startMetrics, MetricsOptions } from './metrics';
+import { startProfiling, ProfilingOptions } from './profiling';
+import { startTracing, TracingOptions } from './tracing';
+
+interface GeneralOptions {
```
seemk, line 21: Metrics and profiling options have a `resource` field, what do you think about adding it to tracing as well? So we can add a common `resource` to `GeneralOptions`

src/start.ts
```diff
@@ -0,0 +1,63 @@
...
+  accessToken: string;
```
seemk, line 22: By reading the types, `accessToken` seems to be required, even though `Partial<Options>` is used in `start`. Are there any downsides to making it optional here?

src/start.ts
```diff
@@ -0,0 +1,63 @@
...
+   * their own default.
+   */
+  endpoint?: string;
+  serviceName: string;
+}
+
+interface Options extends GeneralOptions {
+  metrics: MetricsOptions;
+  profiling: ProfilingOptions;
+  tracing: TracingOptions;
```
seemk, line 34: If one were to enable metrics and profiling, `start` would look like:
```typescript
start({ profiling: {}, metrics: {} });
```

Would it make more sense if it's either boolean or the configuration explicitly in types? :thinking:

docs/advanced-config.md
```diff
@@ -75,3 +75,30 @@ The following config options can be set by passing them as arguments to `startTr
    - `node_version`: `process.versions.node`, e.g. `16.10.0`
 
 \*: Overwritten default value
+
+### Start all
+
+To control all signals with one call `start()` API can be used:
```
seemk, line 81: Is `signal` used elsewhere? Was confused by this word when I first saw it in this context

src/start.ts
Rauno56: That was the intention. It's implemented as such in the logic but forgot the types. Thanks!

docs/advanced-config.md
Rauno56: It's widely used in OpenTelemetry: [Glossary](https://github.com/open-telemetry/opentelemetry-specification/blob/70fecd2dcba505b3ac3a7cb1851f947047743d24/specification/glossary.md#signals), [Overview](https://github.com/open-telemetry/opentelemetry-specification/blob/0c793b450098e8d7a739e575897707f7db08d8ce/specification/overview.md).

Do you know another term that's ever used as a noun to combine "tracing", "metrics", etc?

src/start.ts
Rauno56: So the user could create the Resource themselves before initializing the SDK?
That's the idea in the end. I didn't want to rewrite everything at once. I'll address thsi somehow in another PR, but since the current implementation doesn't take anything away, I'll leave that as-is for now.

src/start.ts
Rauno56: Are there any upsides to making all the properties optional? The type for `start` is still `Partial<GeneralOptions>` and GeneralOptions is used nowhere for now.

docs/advanced-config.md
seemk: It's fine, perhaps make `signals` link to the glossary here

Rauno56:
@seemk, what about the question in the OP :point_up:

src/start.ts
Rauno56: Note: Metrics does not have a resource field.

src/start.ts
```diff
@@ -0,0 +1,84 @@
...
+  tracing: boolean | TracingOptions;
+}
+
+let runningMetrics: ReturnType<typeof startMetrics> | null = null;
+let runningProfiling: ReturnType<typeof startProfiling> | null = null;
+let runningTracing: ReturnType<typeof startTracing> | null = null;
+
+export const start = (options: Partial<Options>) => {
+  if (runningMetrics || runningProfiling || runningTracing) {
+    console.warn('Splunk APM already started');
```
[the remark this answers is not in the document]
Rauno56: Throw!

docs/advanced-config.md
```diff
@@ -75,3 +75,30 @@ The following config options can be set by passing them as arguments to `startTr
    - `node_version`: `process.versions.node`, e.g. `16.10.0`
 
 \*: Overwritten default value
+
+### Start all
```
seemk, line 79: I think this should become the go-to API, perhaps move this to the top and name the section e.g. `Start instrumentation`?

src/start.ts
seemk: The only upside I thought about is the readability/glance value when going over the types. And I might have missed it, but is `GeneralOptions` used anywhere else? Could merge the 2 fields into `Options` itself

docs/advanced-config.md
Rauno56: I agree, but since this feature is additive I don't think we need to do it right now.

src/start.ts
Rauno56: Sure! I had it separate since I anticipated needing to type only general options somewhere. Merging everything into one for now.

PR #479, opened 2022-05-25
feat: throw when unexpected configuration provided

Rauno56:
# Description

I've accountered myself a situation where I think I specify a configuration option, but it does not go into effect because I've mistyped it. This feature catches those in bootup.

## Type of change

# How Has This Been Tested?

- [ ] Tested manually
- [ ] Added automated tests

# Checklist:

### DIFF
--- a/src/tracing/options.ts
+++ b/src/tracing/options.ts
@@ -29,7 +29,7 @@ import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-grpc';
 import { Metadata } from '@grpc/grpc-js';
 import { JaegerExporter } from '@opentelemetry/exporter-jaeger';
 import { detect as detectResource } from '../resource';
-import { defaultServiceName, getEnvBoolean } from '../utils';
+import { deduplicate, defaultServiceName, getEnvBoolean } from '../utils';
 import { NodeTracerConfig } from '@opentelemetry/sdk-trace-node';
 import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
 import { diag, Span, TextMapPropagator } from '@opentelemetry/api';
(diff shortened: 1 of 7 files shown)


src/tracing/options.ts
```diff
@@ -55,18 +55,32 @@ export type CaptureHttpUriParameters = (
...
+  propagatorFactory: PropagatorFactory;
+  serverTimingEnabled: boolean;
   spanExporterFactory: SpanExporterFactory;
   spanProcessorFactory: SpanProcessorFactory;
-  propagatorFactory: PropagatorFactory;
-  captureHttpRequestUriParams: string[] | CaptureHttpUriParameters;
+  tracerConfig: NodeTracerConfig;
 }
 
+export const allowedTracingOptions = [
```
jtmal-signalfx, line 71: you could extract fields specific to tracing to a new type `TracingOptions`, and then this array can be of type `(keyof TracingOptions)[]`, but it's not a big deal

src/utils.ts
```diff
@@ -62,3 +66,29 @@ export function getEnvNumber(key: string, defaultValue: number): number {
 
   return numberValue;
 }
+
+export function deduplicate(arr: string[]) {
+  return [...new Set(arr)];
+}
+
+const formatStringSet = (set: Set<string> | string[]) => {
+  return [...set.values()].map(item => `"${item}"`).join(', ');
```
jtmal-signalfx, line 75: `.values()` will fail in node 8, time to deprecate it? <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/values#browser_compatibility>

src/tracing/options.ts
Rauno56: Yes, the current type works just as well in compile-time, but  I need it in runtime. Do you know a way to compile runtime vars from types?

src/utils.ts
Rauno56: Interesting that the tests don't fail, right? :) The method you're referring to is on an Array. The code uses [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/values#browser_compatibility). So it's all good!


We're deprecating node 8 in 1.0 for different reasons though regardless!

jtmal-signalfx (approved):
If the diag logger is not set, users won't get any information about issues. It's something that we still need to figure out. If diag was printing to stdout by default that would be best, but I think isn't the case.

PR #492, opened 2022-06-07
feat: call shutdown in `stopTracing()`

Rauno56:
# Description

Run `shutdown` in `stopTracing`.

## Type of change

# How Has This Been Tested?

- [ ] Tested manually
- [ ] Added automated tests

# Checklist:


src/tracing/index.ts
```diff
@@ -93,13 +93,21 @@ export function startTracing(opts: Partial<Options> = {}): boolean {
   return true;
 }
 
-export function stopTracing() {
+export async function stopTracing() {
```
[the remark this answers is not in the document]
Rauno56: Thanks for taking a look. We should.

src/tracing/index.ts
```diff
@@ -93,13 +93,24 @@ export function startTracing(opts: Partial<Options> = {}): boolean {
...
+  if (allowDoubleStart) {
+    isStarted = false;
+  }
+  // in reality unregistering is not reliable because of the function pointers
+  // floating around everywhere in the user code already and will lead to
+  // unexpected consequences should it be done more than once. We enable it
+  // mostly for tests.
```
shirhuang, line 103: Curious of this note. Do you think this will be safe to call once in an application that will still be live after calling shutdown() for opentelemetry?

Rauno56: If you do not intend to call `startTracing` later on in the process and are OK with having partially monkey-patched(that's how auto-instrumentations work in Node.js) versions of your dependencies, then you should be totally fine!

Do you have such a use-case?

seemk (commented):
Do we have a test that covers what happens when a custom span is created after shutdown is called? If OTel js provides a no-op tracer provider after shutdown then perhaps we can test that we receive a no-op tracer and a no-op span is created. If that's not possible then perhaps just a check we don't get any runtime errors.

Rauno56:
The API is speced to not fail once everything is shut down, but it's totally worth to test it on our side as well.

seemk (approved):

kumarrishav:
can we get it published today?

seemk:
> can we get it published today?

@kumarrishav 1.1.0 is published now

PR #493, opened 2022-06-10
chore: remove builds and tests on outdated Node.js versions

seemk:
Move prebuilds and tests to at least Node.js 12.13


Rauno56 (approved):
I would've left them until they start causing problems, but totally support removing right away as well.

PR #502, opened 2022-07-04
feat: implement a way to collect unformatted stacktrace lines from profiler

Rauno56:
# Description

For pprof it's useful to start with informatted stacktraces. Implemented a way to export that format instead of parsing already formatted stacktrace strings.

This introduces a lot of code duplication which we can remove once we have implemented the pprof exporter.

Changes to `FindClosestActivation` were due to the compiler complaining(see https://github.com/signalfx/splunk-otel-js/pull/502/commits/4933e40f20d14985e645cc69f846820ccbe16d50) probably because the second occurrence made it so it wasn't inlined anymore. The previous implementation was arguably somewhat more elegant, but the current one removes any confusion for the compiler and the unused `endTime`. If there's a better way to approach this, let me know.

## Type of change

# How Has This Been Tested?

- [x] Added automated tests

# Checklist:

### DIFF
--- a/src/native_ext/profiling.cpp
+++ b/src/native_ext/profiling.cpp
@@ -708,6 +708,136 @@ void ProfilingBuildStacktraces(
   }
 }
 
+v8::Local<v8::Array> makeStackLine(const v8::CpuProfileNode* node) {
+  int32_t itemCount = 0;
+  auto jsResult = Nan::New<v8::Array>();
+
+  const char* rawFunction = node->GetFunctionNameStr();
+  const char* rawFileName = node->GetScriptResourceNameStr();
+
+  if (strlen(rawFunction) == 0) {
+    rawFunction = "anonymous";
+  }
+
+  if (strlen(rawFileName) == 0) {
+    rawFileName = "unknown";
+  }
+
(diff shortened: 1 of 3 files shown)


src/native_ext/profiling.cpp
```diff
@@ -247,29 +247,21 @@ ActivationBin* ProfilingGetActivationBin(Profiling* profiling, int64_t binIndex)
 }
 
 SpanActivation* FindClosestActivation(Profiling* profiling, int64_t ts) {
-  SpanActivation sentinel;
-  sentinel.startTime = std::numeric_limits<int64_t>::min();
-  sentinel.endTime = std::numeric_limits<int64_t>::max();
-  SpanActivation* t = &sentinel;
-
+  SpanActivation* t = nullptr;
```
seemk, line 250: What was your g++ version? The compiler is wrong to produce a warning here as there was a check for the sentinel value before returning. Does it happen in CI as well? This loop is the hottest path when doing span matching

Rauno56: `g++ (GCC) 12.1.1 20220507`.
Yeah, I see that it's checked for, but assumed the compiler didn't catch that.

seemk (approved):
Ideally I'd hoist out the if from the tight loop, but perhaps we can think of something even better to do the matching soon

PR #506, opened 2022-07-13
fix: handle endpoint scheme for profiling exporter

seemk:
Previously we only allowed endpoints without a scheme, but according to the OTel spec, `http` and `https` schemes must be handled.

Allowed values now:
* `host:port` - insecure gRPC credentials
* `http://host:port` - insecure gRPC credentials
* `https://host:port` - secure gRPC credentials,  certs are loaded according to OTLP cert env vars

### DIFF
--- a/src/profiling/OTLPProfilingExporter.ts
+++ b/src/profiling/OTLPProfilingExporter.ts
@@ -16,6 +16,7 @@
 import * as protoLoader from '@grpc/proto-loader';
 import * as grpc from '@grpc/grpc-js';
 import * as path from 'path';
+import * as fs from 'fs';
 import { ProfilingData, ProfilingExporter } from './types';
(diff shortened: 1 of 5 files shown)


src/profiling/OTLPProfilingExporter.ts
```diff
@@ -31,6 +32,48 @@ interface LogsClient extends grpc.Client {
   export: (request: unknown, metadata: grpc.Metadata, callback: Function) => {};
 }
 
+function readContentSync(location: string): Buffer | undefined {
+  try {
+    return fs.readFileSync(path.resolve(process.cwd(), location));
```
Rauno56, line 37: Paths for `fs` are resolved relative to cwd by default.

Rauno56 (approved):
:+1:

PR #514, opened 2022-07-18
test: e2e test enabling profiling

Rauno56:
# Description

Profiling had nothing to smoke test the final packaged version. Building #507 I noticed that I could've shipped a version that breaks prod without catching this in tests.

## Type of change

# How Has This Been Tested?

- [ ] Tested manually
- [x] Added automated tests

# Checklist:

### DIFF
--- a/examples/profiling/README.md
+++ b/examples/profiling/README.md
@@ -1,0 +1,27 @@
+# Profiling Example
+
+> :warning: Profiling is still an experimental feature
+
+This example showcases enabling profiling for Splunk APM. There's no official support for profiling in OTel, so profiling requires working with some Splunk-specific components.
+By default, the example requires the OTel Collector to run with the OTLP receiver listening for logs on `localhost:4317`. To export profiling data to the APM Collector must be used with `splunk_hec` exporter. See [the example collector config](./collector-config.yml).
+
+```shell
+# Replace <...> with the correct values
+export SPLUNK_REALM="<realm>"
+export SPLUNK_ACCESS_TOKEN="<your access token>"
+# Docker Compose configuration has been provided for convenience:
+docker compose up
+```
+
+Then run the script in a separate terminal:
+
+```shell
+# Optional. To set the environment:
+export OTEL_SERVICE_NAME='profiling-example'
+export OTEL_RESOURCE_ATTRIBUTES='deployment.environment=dev'
+export OTEL_LOG_LEVEL='DEBUG'
+# Run the example:
+npm start
+```
+
+The script will then export to collector on shutdown and the collector will take care of transforming the payload into `splunk_hec` format.
--- a/examples/profiling/README.md
+++ b/examples/profiling/README.md
@@ -3,7 +3,7 @@
 > :warning: Profiling is still an experimental feature
 
 This example showcases enabling profiling for Splunk APM. There's no official support for profiling in OTel, so profiling requires working with some Splunk-specific components.
-By default, the example requires the OTel Collector to run with the OTLP receiver listening for logs on `localhost:4317`. To export profiling data to the APM Collector must be used with `splunk_hec` exporter. See [the example collector config](./collector-config.yml).
+By default, the example requires the OTel Collector to run with the OTLP receiver listening for logs on `localhost:4317`. To export profiling data to APM, you must set up `splunk_hec` exporter in the Collector. See [the example collector config](./collector-config.yml).
 
 ```shell
 # Replace <...> with the correct values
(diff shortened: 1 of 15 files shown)


Rauno56:
That's right. 1. to check it's not failing; 2. actually have an example for enabling profiling.

How could we check for profiling data? Afaik `httpsink` only exports traces. [Link](https://github.com/signalfx/splunk-otel-collector/blob/main/internal/exporter/httpsinkexporter/httpsink_exporter.go).

examples/profiling/index.js
```diff
@@ -0,0 +1,38 @@
...
+	},
+});
+
+const doWork = () => {
+	const start = Date.now();
+	while (Date.now() - start < 2000) {}
+};
+
+// setTimeout has to be here because profiling is currently started asyncronously to avoid blocking the runtime.
+// If we didn't we'd run stop before the profiling has started in the background.
```
seemk, line 23: Could make the native init run in the same loop step, don't think the `setImmediate` is important there. Can do it separately tho

examples/profiling/index.js
```diff
@@ -0,0 +1,38 @@
...
+// If OTEL_LOG_LEVEL env var is set, configure logger
+if (process.env.OTEL_LOG_LEVEL) {
+  diag.setLogger(new DiagConsoleLogger(), DiagLogLevel[process.env.OTEL_LOG_LEVEL]);
+}
+
+// Profiling is still experimental and has to be enabled explicitly
+start({
+	profiling: {
+		callstackInterval: 100,
+		collectionDuration: 1_000,
```
seemk, line 13: If it's example code then we're really omitting a lot of information here (service name, resource attributes, endpoint) as they come from the test's env file in a separate directory.

Besides by default we are currently recommended to use a call stack interval of 1000, both of these parameters go into the advanced usage/tweaks category.

examples/profiling/index.js
Rauno56: Will change that separately from adding the example.

examples/profiling/index.js
Rauno56: So what's your suggestion?

Choose to add those options because they are informative(specific to profiling) and make the example run faster. 30s might make the user question whether it's broken or not.

Perhaps I should add a note about the default being recommended for prod?

seemk: Another option would be to move this file to the test dir itself?

Rauno56: Would the motivation mainly be to remove an example that showcases the usage of unusual options or something else? We don't have the general options available in every example anyway. User has to provide a API token at minimum anyway. Resource attributes and endpoint config is shown in other examples - only makes it more confusing to include it in all of them.

We'd lose the example and have to change how the e2e tests are built. Imho not worth it.

Rauno56: The comment about `setTimeout` is suboptimal for now, but it keeps us honest and we'll fix it tomorrow.

seemk: At least a small README would be helpful in this case, as is in https://github.com/signalfx/splunk-otel-js/tree/main/examples/basic

Rauno56: Yes, that makes sense. Added the readme

examples/profiling/README.md
```diff
@@ -0,0 +1,27 @@
+# Profiling Example
+
+> :warning: Profiling is still an experimental feature
+
+This example showcases enabling profiling for Splunk APM. There's no official support for profiling in OTel, so profiling requires working with some Splunk-specific components.
+By default, the example requires the OTel Collector to run with the OTLP receiver listening for logs on `localhost:4317`. To export profiling data to the APM Collector must be used with `splunk_hec` exporter. See [the example collector config](./collector-config.yml).
```
aurbiztondo-splunk, line 6: "To export profiling data to the APM Collector must be used with `splunk_hec` exporter. " is confusing. 

What do we mean, exactly? Is there a comma missing, as in "To export profiling data to APM, the Collector must be used with `splunk_hec` exporter."?

seemk (approved):
Mentioning `SPLUNK_PROFILER_LOGS_ENDPOINT` / endpoint param might be useful

examples/profiling/README.md
Rauno56: Ingest requires `splunk_hec` which can be set up in the collector, so contrary to tracing telemetry, which can be exported directly to the APM, profiling currently requires collector.

Rauno56:
I think we don't need to list every single option. That's what docs are for.

When would user be required to set `SPLUNK_PROFILER_LOGS_ENDPOINT`? The current example happily exports to the default endpoint and if the collector would be available externally, setting the general endpoint env var is sufficient.

examples/profiling/README.md
aurbiztondo-splunk: How about: "To export profiling data to APM, you must set up splunk_hec in the Collector".

Rauno56: Made the change and merged. Thanks.

PR #523, opened 2022-08-01
feat: export pprof (#517)

seemk:
Modification of https://github.com/signalfx/splunk-otel-js/pull/517


jtmal-signalfx (approved):
Looks mostly cosmetic, so I'm approving, if there's something bigger hiding in there that I should take a deeper look at, let me know.

seemk:
I moved it from Rauno's branch so I can edit a bit, for example `source.event.period` should be a label. Basing memory profiling on this branch as well so I can change stuff if necessary.

seemk:
Additional changes according to the [spec](https://github.com/signalfx/gdi-specification/blob/main/specification/semantic_conventions.md#pprof-profileproto-data-format):
* Moved `source.event.period` from `LogRecord` to sample labels
* Serialize unknown lines as -1

Plus
* The values received from native profiler can't be undefined, removed unnecessary function types.
* Native side already checks for empty function names, removed adding of `(anonymous)` due to that

seemk:
Fixed an issue where the native side was generating all identical stacktraces during the collection step. Happened because the stacktraces array was created outside of the stack line creation loop, thus each stacktrace creation overwrote the same array.

PR #524, opened 2022-08-07
feat: memory profiling

seemk:
:warning:  Targeting `pprof-export` branch, will need to be changed after `pprof-export` is merged.

### tldr

* Use `v8::HeapProfiler` to periodically capture an allocation profile along with the samples.
* The allocation samples returned from v8 are "live memory", so we have to keep track which samples were added.
* Samples have a `node_id` which is its node ID in the call graph. To generate a stack trace this node needs to be found from the allocation profile.
* For the finding the whole allocation profile graph is converted to a more convenient format, so when we have a node id, we can do a dictionary lookup and reconstruct the path to a root callgraph node which maps nicely to the pprof location array. The conversion requires generating the whole graph. There is room for improvement here (e.g. caching the node graph and only regenerate it when we find an unknown node_id from the samples), but at the moment it seems to be fast enough.

### Misc

* Removed obsolete Node.js versions (<12) from `prebuild:os` script

### DIFF
--- a/src/native_ext/memory_profiling.cpp
+++ b/src/native_ext/memory_profiling.cpp
@@ -1,0 +1,176 @@
+#include "memory_profiling.h"
+#include <v8-profiler.h>
+#include <vector>
+#include "util/platform.h"
+
+namespace Splunk {
+namespace Profiling {
+
+namespace {
+enum MemoryProfilingStringIndex {
+  V8String_Name,
+  V8String_ScriptName,
+  V8String_LineNumber,
+  V8String_Allocations,
+  V8String_Parent,
+  V8String_MAX
+};
+
+struct BFSNode {
+  BFSNode(v8::AllocationProfile::Node* node, int32_t parentIndex)
+    : node(node), parentIndex(parentIndex) {}
+  v8::AllocationProfile::Node* node;
+  // Parent index in the flattened tree
+  int32_t parentIndex;
+};
+
+struct MemoryProfiling {
+  MemoryProfiling() {
+    stack.reserve(128);
+  }
+  std::vector<BFSNode> stack;
+  bool isRunning = false;
+};
+
+MemoryProfiling* profiling = nullptr;
+
+struct StringStash {
+  v8::Local<v8::String> strings[V8String_MAX];
+};
+
+v8::Local<v8::Object> ToJsHeapNode(v8::AllocationProfile::Node* node, int32_t parentIndex, StringStash* stash) {
+  auto jsNode = Nan::New<v8::Object>();
+  Nan::Set(jsNode, stash->strings[V8String_Name], node->name);
+  Nan::Set(jsNode, stash->strings[V8String_ScriptName], node->script_name);
+  Nan::Set(jsNode, stash->strings[V8String_LineNumber], Nan::New<v8::Integer>(node->line_number));
+  Nan::Set(jsNode, stash->strings[V8String_Parent], Nan::New<v8::Integer>(parentIndex));
+
+  auto jsAllocations = Nan::New<v8::Array>(node->allocations.size());
+  Nan::Set(jsNode, stash->strings[V8String_Allocations], jsAllocations);
+
+  for (size_t allocationIndex = 0; allocationIndex < node->allocations.size(); allocationIndex++) {
+    v8::AllocationProfile::Allocation* allocation = &node->allocations[allocationIndex];
+    Nan::Set(
+      jsAllocations, allocationIndex, Nan::New<v8::Number>(allocation->size * allocation->count));
+  }
+
+  return jsNode;
+}
+
+}
+
+NAN_METHOD(StartMemoryProfiling) {
+  if (!profiling) {
+    profiling = new MemoryProfiling();
+  }
+
+  if (profiling->isRunning) {
+    return;
+  }
+
+  v8::HeapProfiler* profiler = info.GetIsolate()->GetHeapProfiler();
+
+  if (!profiler) {
+    Nan::ThrowError("unable to get heap profiler - isolate not initialized"); 
+    return;
+  }
+
+  int64_t sampleIntervalBytes = 1024 * 256;
+  int32_t maxStackDepth = 256;
+
+  if (info.Length() >= 1 && info[0]->IsObject()) {
+    auto options = Nan::To<v8::Object>(info[0]).ToLocalChecked();
+    auto maybeSampleIntervalBytes =
+      Nan::Get(options, Nan::New("sampleIntervalBytes").ToLocalChecked());
+    if (!maybeSampleIntervalBytes.IsEmpty() && maybeSampleIntervalBytes.ToLocalChecked()->IsNumber()) {
+      sampleIntervalBytes = Nan::To<int64_t>(maybeSampleIntervalBytes.ToLocalChecked()).FromJust();
+    }
+
+    auto maybeMaxStackDepth =
+      Nan::Get(options, Nan::New("maxStackDepth").ToLocalChecked());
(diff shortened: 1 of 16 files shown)


docs/profiling.md
```diff
@@ -27,3 +28,23 @@ startProfiling({
 
 startTracing({ ... });
 ```
+
+
+### Memory profiling
+
+Memory profiling is disabled by default, it can be enabled via the `memoryProfilingEnabled` flag.
```
aurbiztondo-splunk, line 35: Suggestion: Memory profiling is disabled by default. You can enable it via the `memoryProfilingEnabled` flag.

docs/profiling.md
```diff
@@ -27,3 +28,23 @@ startProfiling({
...
+
+Internally the profiler uses V8's sampling heap profiler, where it periodically queries for new allocation samples from the allocation profile.
+
+The [V8 heap profiler's parameters](https://v8.github.io/api/head/classv8_1_1HeapProfiler.html#a6b9450bbf1f4e1a4909df92d4df4a174) can additionally be tuned by an optional `memoryProfilingOptions` configuration field:
```
aurbiztondo-splunk, line 39: Suggestion: You can tune [V8 heap profiler's parameters](https://v8.github.io/api/head/classv8_1_1HeapProfiler.html#a6b9450bbf1f4e1a4909df92d4df4a174) using the `memoryProfilingOptions` configuration field:

aurbiztondo-splunk (approved):

docs/profiling.md
seemk: Thanks! Perhaps you already know, but GitHub lets you add suggestions directly to the comment via a `suggest` code block, so the author can press a button to bring the change in, for example:

```suggestion
Memory profiling is disabled by default. You can enable it via the `memoryProfilingEnabled` flag.
```

src/native_ext/memory_profiling.cpp
```diff
@@ -0,0 +1,219 @@
...
+  stash.strings[V8String_Name] = Nan::New<v8::String>("name").ToLocalChecked();
+  stash.strings[V8String_ScriptName] = Nan::New<v8::String>("scriptName").ToLocalChecked();
+  stash.strings[V8String_LineNumber] = Nan::New<v8::String>("lineNumber").ToLocalChecked();
+  stash.strings[V8String_ParentId] = Nan::New<v8::String>("parentId").ToLocalChecked();
+
+  std::vector<BFSNode>& stack = profiling->stack;
+  stack.clear();
+
+  // Cut off the root node
+  for (size_t i = 0; i < root->children.size(); i++) {
```
Rauno56, line 170: Could we use range-based for loop here and below as well?

src/native_ext/profiling.cpp
```diff
@@ -368,8 +364,11 @@ NAN_METHOD(StartProfiling) {
 
     auto maybeMaxSampleCutoffDelay =
       Nan::Get(options, Nan::New("maxSampleCutoffDelayMicroseconds").ToLocalChecked());
-    if (!maybeMaxSampleCutoffDelay.IsEmpty() && maybeMaxSampleCutoffDelay.ToLocalChecked()->IsNumber()) {
-      int64_t maxSampleCutoffDelayMicros = Nan::To<int64_t>(maybeMaxSampleCutoffDelay.ToLocalChecked()).FromJust();
+    if (
```
Rauno56, line 367: Referencing function arguments is just rough and really calls desperately for something readable. Ofc not a breaker here.

Rauno56 (commented):
Mainly curious questions.

src/native_ext/memory_profiling.cpp
```diff
@@ -0,0 +1,219 @@
...
+    if (!kh_exist(tracking, it)) {
+      continue;
+    }
+
+    if (kh_val(tracking, it) != generation) {
+      kh_del(SampleId, tracking, it);
+    }
+  }
+
+  StringStash stash;
```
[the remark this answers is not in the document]
seemk: Can't create v8 objects when the isolate isn't initialized (i.e. global scope)

src/native_ext/profiling.cpp
```diff
@@ -182,9 +182,11 @@ struct Profiling {
   v8::CpuProfiler* profiler;
   int64_t wallStartTime = 0;
   int64_t startTime = 0;
-  // Maximum offset in nanoseconds from profiling start from which a sample is considered always valid.
+  // Maximum offset in nanoseconds from profiling start from which a sample is considered always
```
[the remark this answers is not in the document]
seemk: This is done by clang format

src/native_ext/memory_profiling.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  std::vector<BFSNode>& stack = profiling->stack;
+  stack.clear();
+
+  // Cut off the root node
+  for (size_t i = 0; i < root->children.size(); i++) {
+    stack.emplace_back(root->children[i], root->node_id);
+  }
+
+  while (!stack.empty()) {
```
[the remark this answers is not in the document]
seemk: The samples are a subset of the call graph nodes and they refer to the nodes in the call graph by ID. The call graph wouldn't exist without samples, so in this case treeMap can be empty if there are no samples.

src/native_ext/memory_profiling.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  v8::AllocationProfile* profile = profiler->GetAllocationProfile();
+
+  if (!profile) {
+    return;
+  }
+
+  auto jsResult = Nan::New<v8::Object>();
+  auto jsSamples = Nan::New<v8::Array>();
+  auto jsNodeTree = Nan::New<v8::Object>();
```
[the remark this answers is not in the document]
seemk: Node ID is an incrementing counter. No idea how large it gets and how many gaps will exist. Just used a denser form :man_shrugging:

src/native_ext/profiling.cpp
seemk: And `Nan` is the simplification of the original API!

src/native_ext/memory_profiling.cpp
seemk: No difference really, but added it

src/native_ext/memory_profiling.cpp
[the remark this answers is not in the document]
Rauno56:
```
const arr = [];
array[5] = "hello";
array[10] = "world";
```

Makes a "holey" array, which is equivalent in terms of the density. I assume calling the same API from the native side behaves the same.

src/native_ext/memory_profiling.cpp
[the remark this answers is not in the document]
Rauno56: Thanks for clarifying it!

src/native_ext/memory_profiling.cpp
Rauno56: That's why I was asking - wondering whether there is a difference and, if not, whether we'd like to do the same thing the same way.

src/native_ext/profiling.cpp
Rauno56: :D

Rauno56 (approved):

src/native_ext/memory_profiling.cpp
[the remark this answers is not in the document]
seemk: Tested it and it actually turns to `DICTIONARY_ELEMENTS` type even with a small app as the node IDs get large, thus it's basically the same as using an object :thinking: And for some reason the average read speed was actually faster when using an object (no idea why) while the write speeds stayed the same.

Issue #546, opened 2022-09-05
@splunk/otel fails to install on centos7


michi88:
Example docker command that fails: 

```
docker run --rm -it centos:7 /bin/bash -c 'curl -sL https://rpm.nodesource.com/setup_14.x | bash - && yum install -y nodejs gcc-c++ make && npm install @splunk/otel'
```

```
$ docker run --rm -it centos:7 /bin/bash -c 'curl -sL https://rpm.nodesource.com/setup_14.x | bash - && yum install -y nodejs gcc-c++ make && npm install @splunk/otel'

## Installing the NodeSource Node.js 14.x repo...


## Inspecting system...

+ rpm -q --whatprovides redhat-release || rpm -q --whatprovides centos-release || rpm -q --whatprovides cloudlinux-release || rpm -q --whatprovides sl-release || rpm -q --whatprovides fedora-release
+ uname -m

## Confirming "el7-x86_64" is supported...

+ curl -sLf -o /dev/null 'https://rpm.nodesource.com/pub_14.x/el/7/x86_64/nodesource-release-el7-1.noarch.rpm'

## Downloading release setup RPM...

+ mktemp
+ curl -sL -o '/tmp/tmp.NOfofFSFX8' 'https://rpm.nodesource.com/pub_14.x/el/7/x86_64/nodesource-release-el7-1.noarch.rpm'

## Installing release setup RPM...

+ rpm -i --nosignature --force '/tmp/tmp.NOfofFSFX8'

## Cleaning up...

+ rm -f '/tmp/tmp.NOfofFSFX8'

## Checking for existing installations...

+ rpm -qa 'node|npm' | grep -v nodesource

## Run `sudo yum install -y nodejs` to install Node.js 14.x and npm.
## You may run dnf if yum is not available:
     sudo dnf install -y nodejs
## You may also need development tools to build native addons:
     sudo yum install gcc-c++ make
## To install the Yarn package manager, run:
     curl -sL https://dl.yarnpkg.com/rpm/yarn.repo | sudo tee /etc/yum.repos.d/yarn.repo
     sudo yum install yarn

Loaded plugins: fastestmirror, ovl
Determining fastest mirrors
 * base: centos.wielun.net
 * extras: centos.wielun.net
 * updates: centos.wielun.net
base                                                                                                                                                                                                                                                                                                                             | 3.6 kB  00:00:00
extras                                                                                                                                                                                                                                                                                                                           | 2.9 kB  00:00:00
[...]

PR #550, opened 2022-09-12
fix: add compilation support for older (<8) GCC versions

seemk:
Older GCC versions don't support `-Wno-cast-function-type` which is needed to include either `v8` or `nan` due to these causing warnings (which we take as errors). Resolved it with pragmas and a compiler version check.

`-Wno-attributes` is used to get rid of unknown (to GCC 4.x) deprecation attributes, e.g. `[[deprecated]]`.

Fixes https://github.com/signalfx/splunk-otel-js/issues/546

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -21,9 +21,7 @@
           "-Werror"
         ],
         "cflags_cc": [
-          "-Wno-unknown-warning",
-          "-Wno-missing-attributes",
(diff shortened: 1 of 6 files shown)


michi88:
Thanks a lot for doing this!

t2t2 (approved):

Rauno56 (approved):

PR #601, opened 2022-11-14
chore: get rid of beachball

seemk:
* Move back to manual CHANGELOG updates.
* Releaser script will copy the changes between current and previous version to the release message as-is.


mhennoch (approved):

jtmal-signalfx (approved):
approved with all my heart

PR #771, opened 2023-08-02
feat: support native builds with Node 20

seemk:
And add Node 20 to testing matrix.


t2t2 (approved):

lbeschastny:
Looks like it's not that straightforward. The build in `1.x` branch didn't work after cherry-picking the changes from master.

And I'm not sure if it's worth investigating and fixing.

PR #804, opened 2023-09-27
fix: compare kafka header keys case insensitively

seemk:
OTel JS defines carrier keys as lowercase. b3 propagator attempts to read lowercase keys, but other SDKs (Java) produce headers with mixed case keys, causing split traces - https://github.com/open-telemetry/opentelemetry-js/issues/3150

This works fine in the case of HTTP, where node's http module lowercases all keys, but kafkajs retains the key casing.

### DIFF
--- a/src/instrumentations/external/kafkajs/propagator.ts
+++ b/src/instrumentations/external/kafkajs/propagator.ts
@@ -22,7 +22,19 @@ adding toString() to make sure string is returned
 */
 export const bufferTextMapGetter: TextMapGetter = {
   get(carrier, key) {
-    return carrier?.[key]?.toString();
+    if (carrier === undefined) {
+      return undefined;
+    }
+
+    const keys = Object.keys(carrier);
+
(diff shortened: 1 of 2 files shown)


src/instrumentations/external/kafkajs/propagator.ts
```diff
@@ -22,7 +22,19 @@ adding toString() to make sure string is returned
 */
 export const bufferTextMapGetter: TextMapGetter = {
   get(carrier, key) {
-    return carrier?.[key]?.toString();
+    if (carrier === undefined) {
```
t2t2, line 25:
```suggestion
    if (!carrier) {
```

Tbh any non-object value is probably unexpected

t2t2 (approved):

src/instrumentations/external/kafkajs/propagator.ts
seemk: :handshake:

PR #815, opened 2023-10-13
Docker multi-arch support

atoulme:
* Add additional platform support for the Docker images
* Allow manual push of images


jvoravong (approved):

seemk (approved):

seemk:
Can't merge due to the unrelated link check failure, could you set `fail: false` in `.github/workflows/check-links.yml`? I'll deal with this on Monday (don't think changelog should be in the link checker anyway) and I can't open a PR myself at the moment as there is no one to approve it atm

PR #892, opened 2024-03-24
Extract configureInstrumentations from tracing

mhennoch:


src/instrumentations/index.ts
```diff
@@ -233,3 +260,82 @@ export function getInstrumentations() {
...
+
+  const startMetricsOptions = Object.assign(
+    pick(restOptions, allowedMetricsOptions),
+    metrics
+  );
+
+  assertNoExtraneousProperties(startMetricsOptions, allowedMetricsOptions);
+  const metricsOptions = setDefaultMetricsOptions(startMetricsOptions);
+
+  configureInstrumentations({
```
seemk, line 335: Would it make sense to split `configureInstrumentations` from this function and rename it just `parseOptions`?

seemk (approved):

PR #985, opened 2024-11-27
feat: change the default sampler to be always on

seemk:
If the sampler is not specified, default to `AlwaysOn`


mhennoch (approved):
LGTM. Thanks for working on this!!

PR #991, opened 2024-12-10
Update ci.yml

breedx-splk:
The actions/upload-artifact and download-artifact are going away in Jan and will break builds. This updates to @v4.


breedx-splk:
Not sure what's up with this. Hopefully the builds aren't impacted next month. 🤷🏻

PR #1077, opened 2025-11-03
Update configs to match gdi-spec changes

breedx-splk:
Aligns with [#353](https://github.com/signalfx/gdi-specification/pull/353) in gdi-spec. 
This should be considered a breaking change for any users currently using snapshots.


seemk (changes requested):
Thanks 🙏

PR #1090, opened 2026-02-25
OpAMP support

seemk:
New env vars:
* `SPLUNK_OPAMP_ENABLED` - defaults to `false`
* `SPLUNK_OPAMP_ENDPOINT` - defaults to `http://localhost:4320/v1/opamp`
* `SPLUNK_OPAMP_POLLING_INTERVAL` - defaults to `30000`

At the moment the client reports status, health and effective config. We haven't fleshed out the effective config completely yet, at the moment it is sending the configuration provided via env vars or YAML - but I think this is wrong. It should take into account programmatic and default values as well. Will iterate on this once it's fleshed out completely.


src/opamp/uuid.ts
```diff
@@ -0,0 +1,18 @@
+import { randomBytes } from 'node:crypto';
```
t2t2, line 17: No header?

t2t2 (approved):

src/opamp/uuid.ts
seemk: Weird the linter doesn't catch it, will debug it seprately. added it manually for now

PR #1106, opened 2026-04-08
Pin all github actions to hash instead of version tag/string

breedx-splk:


.github/workflows/check-links.yml
```diff
@@ -9,15 +9,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Restore lychee cache
-        uses: actions/cache@v4
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
```
Kielek, line 12: I would make manual check and add full tag information (v4.0.0/any real one).
It should be correctly handled by dependabot/renovate) Without this, only hashes will be udpated.

seemk (approved):

PR #1112, opened 2026-04-13
Publishing from GitHub

seemk:
Converted the gitlab release script to github. The existing helper scripts (prepare-release-artifact.mjs) are no longer necessary, but keeping them until the pipeline is up and running.

The first test run will be a dry run.

Releases and tags will be deleted manually.


t2t2 (approved):
let's go

PR #1128, opened 2026-04-23
Replace action with simple gh cli usage

breedx-splk:
We can (and should) use the gh cli to create the release, rather than depending on a 3rd party action for this.


seemk:
Thanks!

seemk (approved):

PR #1130, opened 2026-04-24
Vendor in fastify intrumentation

seemk:
Each time OTel updates we need to wait for fastify instrumentation to do a release else we will bring in multiple `@opentelemetry/instrumentation` packages.


src/instrumentations/external/fastify/NOTICE
t2t2, line 1: empty?

t2t2 (approved):

PR #1142, opened 2026-05-06
httpdc: end span on request close

seemk:
It's possible Node's HTTP library won't emit neither `http.client.request.error` nor `http.client.response.finish`, end the span when the request closes.

Mainly to address Windows flakiness issues.


t2t2 (approved):
🙏

PR #1159, opened 2026-06-03
[OpAMP] Implement gdi-spec effective config changes

t2t2:
Implement effective config specification https://github.com/signalfx/gdi-specification/commit/7886bf4478686df35eb0440df03bb771c7a237e6

<details>
<summary>Summary of pre-PR spec compliance (generated by claude)</summary>

## Effective Environment Config

| ¶ | Requirement | Status |
|---|---|---|
| B1 | Applies only without declarative config | ✅ Gating correct — configuration.ts:708 |
| B2 | File must be named `environment` | ❌ Uses literal `'env'` as map key — OpAMPClient.ts:165 |
| B3 | `text/plain; format=properties; vendor=splunk; v=1.0.0` | ❌ Returns bare `text/plain` — OpAMPClient.ts:63-69 |
| B4 | `KEY=value\n` grammar | ⚠️ Satisfied incidentally — configuration.ts:715 |
| B5 | Exactly the 10 required keys, no extras | ❌ Dumps **all** `SPLUNK_`/`OTEL_` env vars — configuration.ts:712-717. Required keys not guaranteed present; extras forbidden but emitted |
| B6 | Defaults / `null` for unset | ❌ No defaulting — unset vars omitted |
| B7 | Report runtime-effective value (e.g. profiler failed ⇒ `false`) | ❌ Echoes raw env string, no link to running state |
| B8 | Omit multivalued vars | ⚠️ N/A once B5 restricts the set |

## Effective Declarative Config

| ¶ | Requirement | Status |
|---|---|---|
| C1 | Applies only with declarative YAML | ✅ Gating correct — configuration.ts:708-710 |
| C2 | File named after `OTEL_CONFIG_FILE` (incl. path) | ❌ Uses literal `'yaml'`; distro only reads `OTEL_EXPERIMENTAL_CONFIG_FILE` (start.ts:113), not `OTEL_CONFIG_FILE` |
| C3 | `application/yaml; vendor=splunk; v=1.0.0` | ❌ Returns bare `application/yaml` — OpAMPClient.ts:64 |
| C4 | Filtered minimal body | ❌ Sends the **entire raw YAML file** (`rawGlobalConfiguration`) — configuration.ts:709 |
| C5 | Required dot-path fields | ❌ No projection (data exists in `globalConfiguration` but isn't assembled) |
| C6 | Presence-implies-enabled | ❌ Not implemented |
| C7 | Supply defaults even if absent | ❌ Raw passthrough |
| C8 | Evaluate env-var templates | ✅ **Already compliant** — `envSubstitute` at configuration.ts:106-127, verified by effective_config.test.ts:77-90 |
| C9 | Report all exporters per signal | ❌ No extraction |
</details>

Added effective state holder (`src/opamp/effective-state.ts`), to which each component writes the actual values when they've started (including values set via programmatic API). Effective environment config (which is now limited to keys as spec'd) reads all values from there; whilst effective declarative config returns a view filtered as per spec, using the effective state to correct if a feature was requested but didn't actually start (eg. when the native profiler extension was unavailable)

Review notes from claude:

> - **Env vs declarative asymmetry is intentional.** The flat holder can't represent the structure the declarative format needs (multiple exporters per signal — C9, `batch`/`simple`, `otlp_http`/`otlp_grpc`), so the declarative path walks the parsed config; the env format (one endpoint per signal by wire-format definition) reads the holder.
> - **Endpoint resolution mirrors the SDK exactly** (`resolveOtlpExporterUrl` in `src/utils.ts`), verified against the real `OTLPTraceExporter`'s internally-resolved URL: explicit endpoint → signal env var (verbatim) → base env var (path appended) → protocol default.


src/profiling/index.ts
```diff
@@ -122,11 +123,25 @@ export function startProfiling(options: ProfilingOptions) {
   const extension = loadExtension();
 
   if (extension === undefined) {
+    // Profiling was requested but the native extension is unavailable, so
+    // neither the CPU nor the memory profiler actually started. Report both as
+    // disabled (spec B7/C6) — memory profiling cannot run without the
```
seemk: What does spec B7/C6 refer to?

t2t2: References claude used in the generated table in PR description, cleaned up

seemk (approved):

PR #1170, opened 2026-07-01
Add OpAMP remote configuration support (base + profiler)

t2t2:
Add support for controlling profilers (cpu, memory and callgraphs) over OpAMP - turning on/off, interval

gdi-spec: https://github.com/signalfx/gdi-specification/blob/main/specification/opamp_datamodel.md#remote-configuration


src/profiling/ProfilingController.ts
```diff
@@ -0,0 +1,264 @@
...
+      );
+    } else {
+      recordEffectiveState({
+        profilerEnabled: false,
+        memoryProfilerEnabled: false,
+      });
+    }
+  }
+
+  apply(config: RemoteProfilingConfig): Promise<void> {
```
seemk: Would rename this to applyRemoteConfiguration for clarity

src/profiling/ProfilingController.ts
```diff
@@ -0,0 +1,264 @@
...
+    if (enabled) {
+      this._startCpu(
+        this._baseOptions.callstackInterval,
+        this._baseOptions.memoryProfilingEnabled
+      );
+    } else {
+      recordEffectiveState({
+        profilerEnabled: false,
+        memoryProfilerEnabled: false,
+      });
```
seemk, line 72: Isn't this the default state?

src/profiling/ProfilingController.ts
```diff
@@ -0,0 +1,264 @@
...
+    const memoryEnabled = config.memoryProfiler.enabled;
+
+    if (!enabled) {
+      // Surface a memory-without-cpu request as a failure rather than silently
+      // reporting APPLIED while doing nothing: the memory profiler can only run
+      // alongside the CPU profiler in this version.
+      if (memoryEnabled) {
+        throw new Error(
+          'memory_profiler requires cpu_profiler to be enabled; ignoring memory_profiler'
+        );
```
seemk, line 135: Where are these errors surfaced?

src/profiling/ProfilingController.ts
```diff
@@ -0,0 +1,264 @@
...
+      await this._stopCpu();
+    }
+
+    if (!this._startCpu(samplingInterval, memoryEnabled)) {
+      // cpu_profiler was requested but the native profiler did not start (the
+      // extension is unavailable). Surface it as a failure rather than a silent
+      // APPLIED that produces no profiles — mirroring the callgraphs path.
+      throw new Error(
+        'cpu_profiler requested but the profiling extension is unavailable'
+      );
```
seemk, line 171: Again, where are the errors handled / logged?

src/tracing/snapshots/Snapshots.ts
```diff
@@ -75,35 +128,36 @@ export class SnapshotProfiler {
...
       await this._export(profile);
     }, options.collectionIntervalMs);
     this.collectionLoop.unref();
 
-    process.on('exit', () => {
+    this._onExit = () => {
       this.extension.stop(this.profilerHandle);
-    });
+    };
+    process.on('exit', this._onExit);
```
seemk: What's the onExit for?

src/native_ext/profiling.cpp
```diff
@@ -488,6 +492,14 @@ NAN_METHOD(CreateCpuProfiler) {
     return;
   }
 
+  // createCpuProfiler's contract is to allocate a fresh profiler; a duplicate
```
seemk: We can just make createCpuProfiler return the existing profiler handle (maybe worth renaming to getOrCreateCpuProfiler?) and we can use the handle to configure an existing profiler.

src/profiling/ProfilingController.ts
t2t2: Currently no:

https://github.com/signalfx/splunk-otel-js/blob/0180cc6db2a786d3fac8162cb7bb713261232602/src/opamp/effective-state.ts#L53

When effective config is reported, true = on, false = off, undefined = not yet known (report true if configured)

https://github.com/signalfx/splunk-otel-js/blob/0180cc6db2a786d3fac8162cb7bb713261232602/src/configuration.ts#L946-L949

src/profiling/ProfilingController.ts
t2t2: Caught in 
https://github.com/signalfx/splunk-otel-js/blob/0180cc6db2a786d3fac8162cb7bb713261232602/src/opamp/OpAMPClient.ts#L373-L385
-> sent to `_reportRemoteConfigFailure` which logs it with logger.warn + sets it on remote config status on next opamp poll that happens immediately

src/profiling/ProfilingController.ts
t2t2: (same as above)

src/tracing/snapshots/Snapshots.ts
t2t2: To clean it up when `stop()` is called 

https://github.com/signalfx/splunk-otel-js/blob/0180cc6db2a786d3fac8162cb7bb713261232602/src/tracing/snapshots/Snapshots.ts#L176-L183

<details>
<summary>For context this was brought up when I had codex review the code before creating the PR and then fed claude it's findings to confirm:</summary>

Codex review:

> [P2] Stop the pre-registered snapshot profiler (src/start.ts)
>
> This path now creates a SnapshotProfiler even when callgraphs are inactive, but stop() only stops the ProfilingController and never calls stopSnapshotProfiling(). With remote config enabled, a start/stop cycle leaves the snapshot profiler interval and exit listener alive, and a later start overwrites the global profiler while the old timer keeps running. Please track this profiler in RunningState or call stopSnapshotProfiling() during stop() and clear the global profiler after stopping.

Claude:

> The reviewer is right. `stopTracing()` only shuts down the tracer provider — and `SnapshotSpanProcessor.shutdown()` is a no-op. The `SnapshotProfiler` itself (owning `collectionLoop`, the native handle, and a `process.on('exit')` listener) is never stopped. `stopSnapshotProfiling()` exists but nothing calls it, so across `start()`/`stop()` cycles the interval and exit-listener leak and the module-level `profiler` singleton is left dangling.
>
> Let me look at the start.ts imports and the `process.on('exit')` listener:
>
> Read...
>
> Confirmed the bug — and it's slightly broader than stated: `stopSnapshotProfiling()` is never wired into `stop()` for *either* the remote-config path or the regular `SPLUNK_SNAPSHOT_PROFILER_ENABLED` path. Across `start()`/`stop()` cycles the `collectionLoop` interval, the native handle, and the `process.on('exit')` listener all leak, and the module-level `profiler` singleton is left dangling.
>
> Let me make the fix. First, store the exit listener so it can be removed, and reset the singleton on stop:

...

> Summary
>
> Fixed the [P2] finding: the pre-registered (inactive) snapshot profiler — and the regular snapshot-enabled profiler — were never stopped at shutdown, leaking the collection interval, native handle, and a `process.on('exit')` listener across `start()`/`stop()` cycles.
>
> **Changes:**
>
> - src/tracing/snapshots/Snapshots.ts
> - - Made the `exit` listener detachable: stored it in a `_onExit` field instead of an anonymous closure.
> - - `SnapshotProfiler.stop()` now clears the collection loop + stop timeout, **removes the `exit` listener**, and clears in-flight snapshots before draining the final profile.
> - - `stopSnapshotProfiling()` nulls the module singleton *before* awaiting, so a subsequent `start()` registers a fresh profiler and a concurrent start can't adopt a half-stopped one.
> - src/start.ts
> - - Imported `stopSnapshotProfiling` and call it unconditionally in `stop()` (a safe no-op when no profiler is registered), wired into the teardown promise list near the tracing shutdown.
</details>

src/native_ext/profiling.cpp
```diff
@@ -481,17 +485,37 @@ bool CreateCpuProfilingOptions(const Nan::FunctionCallbackInfo<v8::Value> &info,
...
 
-  Profiling *profiling = SetupProfiling(&opts, info.GetIsolate());
+  // Get-or-create by name: reuse a previously-created profiler of the same name
+  // (re-applying the options) instead of allocating a new one, otherwise set up
+  // a fresh profiler. The registry is append-only (stop() only pauses and
+  // leaves the entry in place), so allocating a fresh profiler on a second SDK
+  // start/stop/start cycle would collide with the stopped entry. Reusing lets
+  // the snapshot profiler be re-registered any number of times, and re-applying
+  // the options picks up a changed sampling interval (honored by the next
+  // startCpuProfiler, since v8 bakes it in at start).
```
seemk: I'd get rid of the whole comment to keep signal to noise ratio higher

src/tracing/snapshots/Snapshots.ts
seemk: I just think it is unnecessary, if the process is already exiting why do we need to stop the profiler?

src/profiling/ProfilingController.ts
```diff
@@ -0,0 +1,264 @@
...
+  samplingInterval: number;
+  memoryEnabled: boolean;
+}
+
+// Applies remote profiling configuration at runtime. Owns the runtime profiler
+// handles so the OpAMP client can stay profiling-agnostic, dispatching parsed
+// config here via the `applyRemoteConfig` callback.
+//
+// `apply()` is serialized: each call awaits the previous one so overlapping
+// remote-config pushes cannot interleave stop/start of the native profiler.
```
seemk: `apply` is `applyRemoteConfiguration?

src/tracing/snapshots/Snapshots.ts
```diff
@@ -120,15 +174,84 @@ export class SnapshotProfiler {
...
+  // native start instead, when the collected samples actually use the new rate.
+  setSamplingInterval(samplingIntervalMs: number) {
+    if (this._samplingIntervalMs === samplingIntervalMs) {
+      return;
+    }
+
+    this._samplingIntervalMs = samplingIntervalMs;
+    this.extension.getOrCreateCpuProfiler(
+      nativeSnapshotOptions(samplingIntervalMs)
+    );
```
seemk, line 244: Will the sampling interval ever be changed if the change is done while snapshot profiling is active?

V8 profiler's SetSamplingInterval can't be updated while a profile is in progress (https://v8docs.nodesource.com/node-24.1/d2/d34/classv8_1_1_cpu_profiler.html#aa652c07923bf6e1a4962653cf09dceb1)

`This method must be called when there are no profiles being recorded.` - there's no wording what happens if it is used mid profile.

Have we tested it?

src/tracing/snapshots/Snapshots.ts
t2t2: mm [have to ask past yourself regarding that](https://github.com/signalfx/splunk-otel-js/blame/main/src/tracing/snapshots/Snapshots.ts#L95-L97), but yeah doesn't seem like it actually has a chance to export anything on exit so removing

src/opamp/OpAMPClient.ts
```diff
@@ -205,11 +261,250 @@ export class OpAMPClient {
...
+      },
+      memoryProfiler: {
+        enabled: alwaysOn?.memory_profiler !== undefined,
+      },
+      callgraphs: {
+        enabled: callgraphs !== undefined,
+        samplingInterval:
+          typeof callgraphs?.sampling_interval === 'number'
+            ? callgraphs.sampling_interval
+            : undefined,
```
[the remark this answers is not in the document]
t2t2: Considering the scope creep already in this PR, I think I'd rather keep for later, after it's defined in gdi-spec

seemk (approved):

PR #1173, opened 2026-07-08
chore: upgrade to typescript 7

seemk:
📈 native compilation


seemk:
Will create a new one

